PR #1, opened 2022-02-01
Add new LSIF indexer with testing infrastructure

olafurpg:
- Add indexer implementation in the `src/` directory
- Add config files for linting, formatting and testing
- Add snapshot tests in the `snapshots/` directory


.github/workflows/ci.yml
```diff
@@ -0,0 +1,18 @@
+name: CI
+on:
+  push:
+  pull_request:
```
varungandhi-src: There should be a branch filter here right? It seems wasteful to run for other pushes.

.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+	"semi": false,
+	"trailingComma": "es5",
+	"singleQuote": true,
+	"useTabs": false,
+	"arrowParens": "avoid"
+      }
```
varungandhi-src:
```suggestion
}
```

snapshots/input/multi-project/packages/a/tsconfig.tsbuildinfo
```diff
@@ -0,0 +1 @@
...1],[16,1],[17,1],[18,1],[19,1],[20,1],[21,1],[4,1],[5,1],[25,1],[22,1],[23,1],[24,1],[26,1],[27,1],[28,1],[6,1],[29,1],[30,1],[31,1],[32,1],[7,1],[33,1],[34,1],[35,1],[36,1],[8,1],[41,1],[37,1],[38,1],[39,1],[40,1],[2,1],[42,1],[11,1],[10,1],[48,2],[46,1],[45,1],[51,3],[47,2],[49,4],[50,2],[52,1],[53,1],[105,5],[106,1],[107,6],[108,7],[109,1],[110,1],[54,8],[55,8],[57,9],[58,10],[59,11],[60,12],[61,13],[62,14],[63,15],[64,16],[65,17],[66,18],[67,18],[68,19],[69,20],[70,21],[71,22],[56,1],[103,1],[72,23],[73,24],[74,25],[104,26],[75,27],[76,28],[77,29],[78,30],[79,31],[80,32],[81,33],[82,34],[83,35],[84,36],[85,37],[86,38],[88,39],[87,40],[89,41],[90,42],[91,43],[92,44],[93,45],[94,46],[95,47],[96,48],[97,49],[98,50],[99,51],[100,52],[101,53],[102,54],[111,1],[112,1],[113,55],[43,1],[44,1]],"semanticDiagnosticsPerFile":[1,9,13,12,3,14,15,16,17,18,19,20,21,4,5,25,22,23,24,26,27,28,6,29,30,31,32,7,33,34,35,36,8,41,37,38,39,40,2,42,11,10,48,46,45,51,47,49,50,52,53,105,106,107,108,109,110,54,55,57,58,59,60,61,62,63,64,65,66,67,68,69,70,71,56,103,72,73,74,104,75,76,77,78,79,80,81,82,83,84,85,86,88,87,89,90,91,92,93,94,95,96,97,98,99,100,101,102,111,112,113,43,44]},"version":"4.4.0-beta"}
```
varungandhi-src: Is this a generated file/does it need to be checked in?

snapshots/input/multi-project/dump.svg
```diff
@@ -0,0 +1,300 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
varungandhi-src: Could you provide some context for this SVG and the other one (how they're generated, why they're being checked in etc.)? I understand that they're showing the relations between different character ranges, monikers etc., but I didn't notice any doc comments or code related to these.

snapshots/output/syntax/src/class.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+    public method(methodParam: string): string {
+//         ^^^^^^ definition example 1.0.0 src/class.ts/Class#method().
+//                ^^^^^^^^^^^ definition example 1.0.0 src/class.ts/Class#method().(methodParam)
+      return this.privateMethod(methodParam)
+//                ^^^^^^^^^^^^^ reference example 1.0.0 src/class.ts/Class#privateMethod().
+//                              ^^^^^^^^^^^ reference example 1.0.0 src/class.ts/Class#method().(methodParam)
+    }
+    public static staticMethod(methodParam: string): string {
+//                ^^^^^^^^^^^^ definition example 1.0.0 src/class.ts/Class#staticMethod().
```
varungandhi-src: TypeScript allows a class to have an instance method with the same name and signature as a static method; if that happens, we'll run into an ambiguity because the syntax is identical (`Class#method()`), right?

snapshots/output/multi-project/src/example.ts
```diff
@@ -0,0 +1,9 @@
+  export function hello() {
+//                ^^^^^ definition example 1.0.0 src/example.ts/hello().
+    const a = 'a'
+//        ^ definition local 2
```
varungandhi-src, line 4: Why is this `local 2` instead of `local 0`? It looks like a new `Visitor` is created per file, so the counter should restart from 0, right?

src/Indexer.ts
```diff
@@ -0,0 +1,73 @@
...
+    this.checker = this.program.getTypeChecker()
+    this.packages = new Packages(options.project)
+  }
+  public index(): void {
+    this.options.writeIndex(
+      new lsif_typed.Index({
+        metadata: new lsif_typed.Metadata({
+          project_root: url.pathToFileURL(this.options.projectRoot).toString(),
+          tool_info: new lsif_typed.ToolInfo({
+            name: 'lsif-node',
```
varungandhi-src: Should we use `lsif-typescript` here?

src/Indexer.ts
```diff
@@ -0,0 +1,73 @@
...
+import { LsifSymbol } from './LsifSymbol'
+import { Options, lsif_typed } from './main'
+import { Packages } from './Packages'
+import { Visitor } from './Visitor'
+
+export class Indexer {
+  public options: Options
+  public program: ts.Program
+  public checker: ts.TypeChecker
+  public symbolsCache: Map<ts.Node, LsifSymbol> = new Map()
```
varungandhi-src: Nit: cache already conveys a collection, so I think we should name this `symbolCache` instead.

src/Indexer.ts
```diff
@@ -0,0 +1,73 @@
...
+  public packages: Packages
```
varungandhi-src: Should all of these fields be `private`? I don't see them being used outside this file.

src/Indexer.ts
```diff
@@ -0,0 +1,73 @@
...
+      })
+    )
+    const sourceFiles = this.program.getSourceFiles()
+    if (sourceFiles.length === 0) {
+      throw new Error('No source files')
+    }
+    // Visit every sourceFile in the program
+    for (const sourceFile of this.program.getSourceFiles()) {
+      const includes = this.config.fileNames.includes(sourceFile.fileName)
+      if (includes) {
```
varungandhi-src: Nit: Using an early `continue` here would reduce nesting. Similarly for `visitor.document.occurrences.length > 0` later in this function.

src/Descriptor.ts
```diff
@@ -0,0 +1,62 @@
...
+  Package,
+  Type,
+  Term,
+  Meta,
+  Method,
+  Parameter,
+  TypeParameter,
+}
+
+export class Descriptor {
```
varungandhi-src: Short doc strings for this type and the `syntax()` method? (I'm wondering if it would be more idiomatic to rename the `syntax()` method to `toString()`.)

src/Indexer.ts
varungandhi-src: And maybe we should bump the version down to `0.1.0` to indicate that this isn't stable yet?

src/Indexer.ts
```diff
@@ -0,0 +1,73 @@
...
+            name: 'lsif-node',
+            version: '1.0.0',
+            arguments: [],
+          }),
+        }),
+      })
+    )
+    const sourceFiles = this.program.getSourceFiles()
+    if (sourceFiles.length === 0) {
+      throw new Error('No source files')
```
varungandhi-src: Could you please add some context to this error (e.g. `projectRoot`?).

src/Input.ts
```diff
@@ -0,0 +1,31 @@
...
+  public lines: string[]
+  constructor(public readonly path: string, public readonly text: string) {
+    this.lines = text.split('\n')
+  }
+
+  public static fromFile(path: string): Input {
+    // eslint-disable-next-line no-sync
+    return new Input(path, fs.readFileSync(path).toString())
+  }
+  public format(range: Range, diagnostic?: string): string {
```
varungandhi-src, line 27: Short doc string? Bit hard to figure out what exactly this is trying to do based on the method body. Of course, it's not that difficult, but would be nice to have a short summary.

src/LsifSymbol.ts
```diff
@@ -0,0 +1,37 @@
...
+  private constructor(public readonly value: string) {}
+  public isEmpty(): boolean {
+    return this.value === ''
+  }
+  public isLocal(): boolean {
+    return this.value.startsWith('local ')
+  }
+  public isEmptyOrLocal(): boolean {
+    return this.isEmpty() || this.isLocal()
+  }
```
varungandhi-src: Nit: Should this be inline? It seems to be used only in one place right now.

src/LsifSymbol.ts
```diff
@@ -0,0 +1,37 @@
...
+  public static sourceFile(
+    package_: LsifSymbol,
+    relativePath: string[]
+  ): LsifSymbol {
+    let symbol = package_
+    for (const part of relativePath) {
+      symbol = LsifSymbol.global(symbol, Descriptor.package(part))
+    }
+    return symbol
+  }
```
varungandhi-src: It looks like this method is unused... Could you turn on warnings (and make it an error in CI) for unused variables?

src/LsifSymbol.ts
```diff
@@ -0,0 +1,37 @@
...
+  public static package(name: string, version: string): LsifSymbol {
+    return new LsifSymbol(`lsif-node npm ${name} ${version} `)
```
varungandhi-src:
```suggestion
    return new LsifSymbol(`lsif-typescript package ${name} ${version} `)
```

src/Packages.ts
```diff
@@ -0,0 +1,66 @@
...
+import { LsifSymbol } from './LsifSymbol'
+
+export class Packages {
+  constructor(public readonly projectRoot: string) {}
+  private cache: Map<string, LsifSymbol> = new Map()
+  public symbol(filePath: string): LsifSymbol {
+    if (path.normalize(filePath) !== filePath) {
+      throw new Error('BOOM FIX THIS')
+    }
+    // TODO: handle end case.
```
varungandhi-src: Is this `TODO` for the `BOOM` line or something else? (If so, would be good to move it the `TODO:` into the `Error`)

src/Packages.ts
```diff
@@ -0,0 +1,66 @@
...
+    const fromCache = this.cache.get(filePath)
+    if (fromCache) {
+      return fromCache
+    }
+    const packageJsonPath = path.join(filePath, 'package.json')
+    try {
+      if (
+        // eslint-disable-next-line no-sync
```
varungandhi-src: Should we turn off this lint altogether? It seems needlessly tedious to turn it off every time you want to do a synchronous I/O operation.

src/Position.ts
```diff
@@ -0,0 +1,12 @@
+export class Position {
+  constructor(
+    public readonly line: number,
+    public readonly character: number
+  ) {}
```
varungandhi-src, line 9: Would be useful to add a doc comment that these are 0-based.

src/Visitor.ts
```diff
@@ -0,0 +1,234 @@
...
+import { Counter } from './Counter'
+import { Descriptor } from './Descriptor'
+import { Input } from './Input'
+import * as lsif from './lsif'
+import { LsifSymbol } from './LsifSymbol'
+import { lsif_typed } from './main'
+import { Packages } from './Packages'
+import { Range } from './Range'
+
+export class Visitor {
```
varungandhi-src: WDYT about giving this type a more descriptive name? E.g. we could rename `Indexer` to `ProjectIndexer`, and call this `FileIndexer`.

src/Visitor.ts
```diff
@@ -0,0 +1,234 @@
...
+    if (ts.isIdentifier(node)) {
+      const sym = this.checker.getSymbolAtLocation(node)
+      if (sym) {
+        this.visitIdentifier(node, sym)
+      }
+    }
+    ts.forEachChild(node, node => this.visit(node))
+  }
+
+  private visitIdentifier(node: ts.Identifier, sym: ts.Symbol): void {
```
varungandhi-src: Nit: rename this parameter to `identifier`? It would make the code below more clearer, and also be less confusing since other methods use `node: ts.Node` whereas this has type `ts.Identifier`.

src/Visitor.ts
```diff
@@ -0,0 +1,234 @@
...
+    }
+    for (const declaration of sym?.declarations || []) {
+      const lsifSymbol = this.lsifSymbol(declaration)
+      this.document.occurrences.push(
+        new lsif.lib.codeintel.lsif_typed.Occurrence({
+          range,
+          symbol: lsifSymbol.value,
+          symbol_roles: role,
+        })
+      )
```
varungandhi-src: Is it possible for `lsifSymbol` here to be empty? If not, would be good to some check + assert or check + emit warning. If it's possible, does it make sense to still add that occurrence? (Similarly for the other place where there's a `this.lsifSymbol` call followed by an `occurrences.push` call.)

src/Visitor.ts
```diff
@@ -0,0 +1,234 @@
...
+      if (isDefinition) {
+        this.addSymbolInformation(node, sym, lsifSymbol)
+        this.handleShorthandPropertyDefinition(declaration, range)
+      }
+    }
+  }
+
+  private handleShorthandPropertyDefinition(
```
varungandhi-src: If possible, it would be useful to have a code example here of what ends up happening if this special handling isn't present. Alternately, the information could be added in the test case for this function (so it's visible in context without needing to write a separate code example), and you can mention the name of that test case here.

src/Visitor.ts
```diff
@@ -0,0 +1,234 @@
...
+    declaration: ts.Node,
+    range: number[]
+  ): void {
+    if (declaration.kind === ts.SyntaxKind.ShorthandPropertyAssignment) {
+      const valueSymbol =
+        this.checker.getShorthandAssignmentValueSymbol(declaration)
+      if (valueSymbol) {
```
varungandhi-src: Nit: early returns would make the code a little cleaner IMO.

src/Visitor.ts
```diff
@@ -0,0 +1,234 @@
...
+      for (const declaration of tpe.symbol?.declarations || []) {
+        return this.lsifSymbol(declaration)
+      }
+    }
+
+    const desc = this.descriptor(node)
+    if (desc) {
+      return this.cached(node, LsifSymbol.global(owner, desc))
+    }
+    return this.newLocalSymbol(node)
```
varungandhi-src: I don't fully understand this last return. I see it being hit in a few cases when the tests are run:

```
hit last code path for a with parent { a }
  hit last code path for (param: string) => string with parent methodSignature2: (param: string) => string
hit last code path for (param: string): string => {
      return param
    } with parent methodSignature2: (param: string): string => {
      return param
    }
```

Why are these cases not being caught earlier with the `owner.isEmptyOrLocal()` check?

src/main.test.ts
```diff
@@ -0,0 +1,136 @@
...
+          index.metadata = partialIndex.metadata
+        }
+        for (const document of partialIndex.documents) {
+          index.documents.push(document)
+        }
+      },
+    })
+    // eslint-disable-next-line no-sync
+    fs.writeFileSync(
+      path.join(projectRoot, 'dump.lsif-typed'),
```
varungandhi-src: This generates `dump.lsif-typed` files under `snapshots/input` (instead of in the `snapshots/output/` directory) when you run tests, which feels a little odd. Intentional or accidental?

src/lsif.ts
```diff
@@ -0,0 +1,1537 @@
+/**
```
varungandhi-src, line 1: You can [create a `.gitattributes` file](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github) so that this is ignored in the future, even if there are small changes (right now, GitHub hid the changes because there was a large diff).

src/Visitor.ts
```diff
@@ -0,0 +1,234 @@
...
+      if (name) {
+        return Descriptor.type(name)
+      }
+    }
+    if (
+      ts.isFunctionDeclaration(node) ||
+      ts.isMethodSignature(node) ||
+      ts.isMethodDeclaration(node)
+    ) {
+      return Descriptor.method(node.name?.getText() || 'boom', '')
```
varungandhi-src:
```suggestion
      const name = node.name?.getText()
      if (name) {
        return Descriptor.method(name, '')
      }
```

src/Descriptor.ts
```diff
@@ -0,0 +1,62 @@
...
+  public static type(name: string): Descriptor {
+    return new Descriptor(name, DescriptorKind.Type)
+  }
+  public static term(name: string): Descriptor {
+    return new Descriptor(name, DescriptorKind.Term)
+  }
+  public static meta(name: string): Descriptor {
+    return new Descriptor(name, DescriptorKind.Meta)
+  }
+  public static method(name: string, disambiguator: string): Descriptor {
```
varungandhi-src: Maybe remove the `disambiguator` for now? I'm assuming it's there for method overloading -- we can re-add it later when we add support for method overloading. Right now, it seems a little strange because both call-sites are passing in `''`.

src/Descriptor.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export class Descriptor {
+  private constructor(
+    public readonly name: string,
+    public readonly kind: DescriptorKind,
+    public readonly disambiguator?: string
+  ) {}
+  public static package(name: string): Descriptor {
+    return new Descriptor(name, DescriptorKind.Package)
```
varungandhi-src: Should this be `module` (or `module_`) instead of `package`? (And similarly for `DescriptorKind.Package`) It looks like this is being used for modules, so it's a little confusing because there is another `static package` which is indeed being used for a package (there's a version involved).

src/Indexer.ts
```diff
@@ -0,0 +1,73 @@
...
+        const input = new Input(sourceFile.fileName, sourceFile.getText())
+        const visitor = new Visitor(
+          this.checker,
+          input,
+          document,
+          this.symbolsCache,
+          this.packages,
+          sourceFile
+        )
+        visitor.index()
```
varungandhi-src: Should this `visitor.index()` call be in a try-catch block? (Maybe we log an error and keep going?) Otherwise, if this throws an error, it will stop indexing the entire package, which may be undesirable (although, I do understand if we want to fail fast while we are doing the bring up).

src/main.ts
```diff
@@ -0,0 +1,178 @@
...
+import * as yargs from 'yargs'
+
+import { Indexer } from './Indexer'
+import * as lsif from './lsif'
+
+export const lsif_typed = lsif.lib.codeintel.lsif_typed
+
+export interface Options {
+  projectRoot: string
+  project: string
```
varungandhi-src: Should we call these `workspaceRoot` and `projectRoot` respectively (or something more distinct)? It seems a little confusing to have `project` in the name for both of them.

src/main.test.ts
```diff
@@ -0,0 +1,136 @@
...
+      const relativeToInputDirectory = path.relative(inputDirectory, inputPath)
+      const outputPath = path.resolve(outputDirectory, relativeToInputDirectory)
+      // eslint-disable-next-line no-sync
+      const expected: string = fs.existsSync(outputPath)
+        ? // eslint-disable-next-line no-sync
+          fs.readFileSync(outputPath).toString()
+        : ''
+      const input = Input.fromFile(inputPath)
+      const obtained = formatSnapshot(input, document)
+      if (obtained !== expected) {
```
varungandhi-src: Early `continue`?

src/main.ts
```diff
@@ -0,0 +1,178 @@
...
+        yargs.positional('project', {
+          type: 'string',
+          default: '.',
+          describe: 'the directory to index',
+        })
+        yargs.option('useYarnWorkspaces', {
+          type: 'boolean',
+          default: 'false',
+          describe: 'whether to index all yarn workspaces',
+        })
```
varungandhi-src: "use" doesn't seem appropriate here. There's already the verb `index` in the doc comment, how about reusing that, so something like `indexAllWorkspaces`/`indexYarnWorkspaces`/`indexAllYarnWorkspaces`.

src/main.ts
```diff
@@ -0,0 +1,178 @@
...
+    throw new Error(
+      ts.formatDiagnostics(result.errors, ts.createCompilerHost({}))
+    )
+  }
+  return result
+}
+
+function defaultCompilerOptions(configFileName?: string): ts.CompilerOptions {
+  const options: ts.CompilerOptions =
+    configFileName && path.basename(configFileName) === 'jsconfig.json'
```
varungandhi-src, line 177: Typo?
```suggestion
    configFileName && path.basename(configFileName) === 'tsconfig.json'
```

.github/workflows/ci.yml
olafurpg: I updated the push filter to only run on the main branch.

snapshots/output/multi-project/src/example.ts
olafurpg: The TypeScript compiler has symbols that have no occurrences. For example, the block `{ ... }` is the parent symbol of `const a` that gets assigned `local 1` or `local 2`

snapshots/output/syntax/src/class.ts
olafurpg: Correct. We currently don't handle overloaded methods, it has a checkbox in the umbrella ticket https://github.com/sourcegraph/sourcegraph/issues/29967

Method symbols have a "disambiguator" part that we can use to distinguish between overloaded methods. See the "Disambiguator" subsection here https://scalameta.org/docs/semanticdb/specification.html#symbol-2 for how this is handled for Java.

snapshots/input/multi-project/dump.svg
olafurpg: I added this file by accident, removed it

snapshots/input/multi-project/packages/a/tsconfig.tsbuildinfo
olafurpg: This was was accidentally added, removed now

src/Descriptor.ts
olafurpg: I refactored this entire file to reuse the built-in `Descriptor` message from lsif.proto. It should be fairly self-documenting now, these are just helper methods to construct descriptors

olafurpg (commented):
@varungandhi-src Thank you for the fantastic code review! I incorporated almost all of the suggestion and the resulting code is a lot cleaner. I'm sorry for the super slow response on this.

I resolved most of the comments without replying because the GitHub PR UI didn't allow me to reply from the "Conversation" view, and the "Files" view was not loading for me.

I have a few more things to address before this is ready for merge.

src/main.ts
olafurpg: Added a comment, jsconfig.json is an actual thing https://sourcegraph.com/search?q=context:global+file:jsconfig.json&patternType=literal

This code was adapted from the original lsif-node implementation.

src/Visitor.ts
olafurpg: This case happens for local symbols that have a global owner. For example, blocks `{}` have a symbol encoding in the compiler and they can be direct children of exported functions.

olafurpg: I added a comment in the code to elaborate on this case.

varungandhi-src (approved):

src/Descriptor.ts
olafurpg: This is a discussion for LSIF Typed, which defines the "package" suffix. The name "package" comes from SemanticDB, which refers to Java/Scala packages. I'd be open to rename it into something else like a namespace (less so 'module')

I opened an issue to continue this conversation https://github.com/sourcegraph/sourcegraph/issues/31855

PR #22, opened 2022-03-02
Handle named import clauses

olafurpg:
Previously, lsif-typescript emitted an incorrect symbol for named import
clauses like `import React from 'react'`. Now, we handle this case
differently from anonymous import clauses such as `import { React } from 'react'`

### Test plan

Snapshot tests in this PR


snapshots/output/react/src/LoaderInput.tsx
```diff
@@ -0,0 +1,38 @@
...
+//                                                                                 ^^^^^^^^^ reference local 6
+      <div className="hello">
+//     ^^^ reference @types/react 17.0.0 `index.d.ts`/global/JSX/IntrinsicElements#div.
+//         ^^^^^^^^^ reference local 11
+          {children}
+//         ^^^^^^^^ reference local 5
+          {loading && <p className="spinner">spinner</p>}
+//         ^^^^^^^ reference local 4
+//                     ^ reference @types/react 17.0.0 `index.d.ts`/global/JSX/IntrinsicElements#p.
+//                       ^^^^^^^^^ reference local 17
```
olafurpg: This should not be a local symbol, it's a public property in the react codebase. I'm investigating why

Issue #46, opened 2022-03-18
Incorrect Go to Def for React element (TSX)


varungandhi-src:
https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/web/src/enterprise/batches/detail/changesets/BatchChangeChangesets.tsx?L263

Go to Definition for `ChangesetFilterRow` takes you to `interface Step`.

olafurpg:
I'm able to reproduce. I just snapshotted the LSIF to see what symbols are getting generated.

```
cd sourcegraph
lsif-typescript index --yarn-workspaces
go build github.com/sourcegraph/sourcegraph/lib/codeintel/tools/lsif-typed
lsif-typed dump.lsif-typed > dump.lsif
lsif-java snapshot-lsif
```

At call-site it references `React/FunctionComponent#`. I recall seeing similar behavior with tsserver via VS Code
```ts
                <ChangesetFilterRow
//               ^^^^^^^^^^^^^^^^^^ reference lsif-typescript npm @types/react 17.0.0 `index.d.ts`/React/FunctionComponent#
```

At definition-site, it weirdly enough defines that same symbol
```ts
interface Step {
//        ^^^^ definition lsif-typescript npm @types/react 17.0.0 `index.d.ts`/React/FunctionComponent# Step
//        ^^^^ definition lsif-typescript npm @sourcegraph/web 1.10.1 src/auth/`PostSignUpPage.tsx`/Step# Step
```

The type of `ChangeFilterRow` is `FunctionComponent<ChangeFilterRowProps>`
```ts
export const ChangesetFilterRow: React.FunctionComponent<ChangesetFilterRowProps> = ({
//           ^^^^^^^^^^^^^^^^^^ definition lsif-typescript npm @sourcegraph/web 1.10.1 src/enterprise/batches/detail/changesets/`ChangesetFilterRow.tsx`/ChangesetFilterRow. FunctionComponent<ChangesetFilterRowProps>
```

Based on that observation, we're using the symbol of the type instead of the symbol itself. This might be happening around how we de-alias import specifiers, which are local symbols by default but lsif-typescript takes the type of the import node.

Fixing the definition site is probably the best first-step towards improving the situation, I have no clue why that definition is getting emitted.

varungandhi-src:
Minimal example:

```
import React from 'react'

import { StepN } from './Steps'

interface Step {
  content: React.ReactElement
}
```

This gives 2 identifiers as you'd expect:

[...]

PR #71, opened 2022-04-04
De-alias symbols similar to tsserver's go-to-definition.

varungandhi-src:
Fixes issue #46.

In the first commit, I've added a generated snapshot to show that we add the
incorrect reference. The fix is implemented in the second commit, where I'm using
logic from tsserver to make sure we de-alias symbols (which appear in the
presence of imports). So the generated snapshots now have the right references.

I've left some TODOs and FIXMEs since there are some edge cases around
JS files and dynamic imports which are handled by tsserver, and handling those
required inlining a bunch of more code. It wasn't super clear to me how all
those edge cases should be tested, so I've avoided inlining those.
I will file a separate follow-up issue to address those TODOs.

### Test plan

Added snapshot tests.

### DIFF
--- a/src/TypeScriptInternal.ts
+++ b/src/TypeScriptInternal.ts
@@ -1,0 +1,38 @@
+import * as ts from 'typescript'
+
+// Functions in this file are based directly off corresponding functions
+// in the TypeScript codebase.
+
+export function shouldSkipAlias(node: ts.Node): boolean {
+  switch (node.kind) {
+    case ts.SyntaxKind.ImportClause:
+    case ts.SyntaxKind.ImportEqualsDeclaration:
+      // TODO: How do we test this code path?
+      return true;
+    case ts.SyntaxKind.ImportSpecifier:
+      return node.parent.kind === ts.SyntaxKind.NamedImports;
+    case ts.SyntaxKind.BindingElement:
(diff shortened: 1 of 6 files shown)


src/TypeScriptInternal.ts
```diff
@@ -0,0 +1,38 @@
...
+  switch (node.kind) {
+    case ts.SyntaxKind.ImportClause:
+    case ts.SyntaxKind.ImportEqualsDeclaration:
+      // TODO: How do we test this code path?
+      return true;
+    case ts.SyntaxKind.ImportSpecifier:
+      return node.parent.kind === ts.SyntaxKind.NamedImports;
+    case ts.SyntaxKind.BindingElement:
+    case ts.SyntaxKind.VariableDeclaration:
+      // TODO: How do we test this code path?
```
Strum355, line 16: Able to put just a standard JS file alongside the other sources?

varungandhi-src: That's not enough, we need to somehow create a BindingElement or VariableDeclaration that is also an alias and I don't know how to do that, because the different `SymbolFlag` values and syntax node types in the TypeScript compiler aren't documented.

Strum355 (approved):

PR #73, opened 2022-04-11
Add docs for working on lsif-typescript.

varungandhi-src:
### Test plan

n/a, only adding docs

### DIFF
--- a/Development.md
+++ b/Development.md
@@ -1,0 +1,44 @@
+# Developing lsif-typescript
+
+## References
+
+- VS Code is a good reference point for the "correct" behavior
+  for code navigation functionality.
+- Keep a local checkout of Microsoft/TypeScript for checking out code examples.
+  Most TypeScript APIs are lacking in documentation,
+  so reading the code is the best way to understand how things work.
+
+## Running tests
+
+```sh
+npm run test
+```
+
+```
+npm run update-snapshots
+```
+
+Generate snapshots and update.
+
+## Snapshotting arbitrary projects
+
+```sh
+npm run build # build lsif-typescript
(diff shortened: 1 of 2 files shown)


Development.md
```diff
@@ -0,0 +1,44 @@
+# Developing lsif-typescript
+
+## References
+
+- VS Code is a good reference point for the "correct" behavior
+  for code navigation functionality.
+- Keep a local checkout of Microsoft/TypeScript for checking out code examples.
```
Strum355, line 7: Not recommending they use Sourcegraph to browse it :smh:

varungandhi-src: Sourcegraph isn't as fast as ripgrep running on an SSD (few things are). 👁️

Development.md
```diff
@@ -0,0 +1,44 @@
...
+  so reading the code is the best way to understand how things work.
+
+## Running tests
+
+```sh
+npm run test
+```
+
+```
+npm run update-snapshots
```
Strum355, line 18: Does this require having lsif-java installed?

varungandhi-src: No, we have this logic in the test code. https://sourcegraph.com/github.com/sourcegraph/lsif-typescript/-/blob/src/main.test.ts?L101:10

Strum355: Are there any other environment requirements we should document? Specific node version etc etc

Strum355 (approved):

Development.md
varungandhi-src: I don't think we need a specific node version, but we can add that later as a config file if needed (vs just having it in the docs).

PR #80, opened 2022-04-13
Update ChangeLog and bump version for releasing 0.1.15.

varungandhi-src:
### Test plan

Ran automated tests.


varungandhi-src:
I've updated the release process so that we can publish releases without requiring a push to the protected main branch from a bot (this is not possible to do from a GitHub Action, and we want to maintain branch protection for SOC2).

bobheadxi (approved):
🚀 🚢 :shipit:

Issue #81, opened 2022-04-13
Auto-indexing for pure JS packages is broken


varungandhi-src:
The `--infer-tsconfig` flag is used by auto-indexing, but the current implementation emits an empty `{}`, which doesn't work. We need to at least have `{"compilerOptions":{"allowJs":true}}`.

PR #83, opened 2022-04-13
fix: Use allowJS in --infer-tsconfig.

varungandhi-src:
lsif-node uses allowJS; however, that was skipped for some reason
in lsif-typescript. Without allowJS, JS files will be excluded
by the TypeScript compiler.

The test is intended for future iteration; and the correctness
of the results doesn't really matter for this PR. The key part
is that we are able to generate LSIF dumps properly for JS.

Fixes issue #81.

### Test plan

Added automated test. Manually tested with [d3-shape](https://github.com/d3/d3-shape).

### DIFF
--- a/snapshots/input/pure-js/package.json
+++ b/snapshots/input/pure-js/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "pure-js",
+  "version": "1.0.0",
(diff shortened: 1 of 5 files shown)


efritz (approved):
image

PR #90, opened 2022-04-21
Update `--infer-tsconfig` to skip the `allowJs` option when there exist TS files

### LINKED ISSUE #87: Update `--inferTSConfig` to dynamically add `--allowJs` based on presence of TS/JS files
It would also be nice to support `lsif-typescript index . -- --allowJs`

olafurpg:
Fixes #87. Previously, we always included the `allowJs` option when
using `--infer-tsconfig`. However, this behavior was undesirable because
TypeScript package repos include a large amount of generated JavaScript
code that should not be indexed.  This commit changes the behavior so
that we only use `allowJs` when the project doesn't have any TypeScript
files.

For example, the `npm/typescript` package (for the TypeScript compiler
itself) times out when we index with `allowJs` while it completes in
a few seconds without `allowJs`.  It's OK to skip the JavaScript files
since the TypeScript files include the definitions for the JavaScript
code.

### Test plan

Added new `inferTsConfig.test.ts` file.

### DIFF
--- a/src/inferTsConfig.test.ts
+++ b/src/inferTsConfig.test.ts
@@ -1,0 +1,21 @@
+import { join } from 'path'
+import * as process from 'process'
+
+import { test } from 'uvu'
+
+import { allowJsConfig, inferTsConfig, noJsConfig } from './inferTsConfig'
+
+const inputDirectory = join(process.cwd(), 'snapshots', 'inferTsConfig')
+
+function checkDirectory(name: string, expected: string): void {
+  test(name, () => {
+    const directory = join(inputDirectory, name)
+    const obtained = inferTsConfig(directory)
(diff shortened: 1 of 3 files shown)


olafurpg:
Gonna go ahead and merge this since @varungandhi-src is on sick leave today and tomorrow. Happy to address review comments after merging.

varungandhi-src (commented):
LGTM; there is already a snapshot test for `--infer-tsconfig`. Should that be tweaked to use this/removed/or be kept as is?

olafurpg:
The existing snapshot test can be kept as is. I covers the use-case where we want a pure JS project without tsconfig to index. It was probably overkill to add dedicated tests for this but I liked having the option to tweak this behavior in isolation down the road.

PR #95, opened 2022-04-28
Correctly emit definition for enum members.

### LINKED ISSUE #93: Add support for enum members
It looks like precise code intel for enum members doesn't work at all currently. ([code](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/search-ui/src/input/extensions/completion.ts?L33-48))

Screenshot showing search-based intel for enum member

varungandhi-src:
Fixes #93.

### Test plan

Updated existing test.


olafurpg (approved):
Woah! Nice find 👍

PR #96, opened 2022-04-28
Fix the Prettier CI check

olafurpg:
Previously, the CI was running prettier and only listing to stdout what files are incorrectly formatted, without failing the CI job with a non-zero exit code.

### Test plan

Watch the CI fail on the first commit (before formatting) and then succeed on the second commit (after formatting).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## LSIF Typescript
 
 ### v0.1.16 (2022/04/26 02:07 +00:00)
+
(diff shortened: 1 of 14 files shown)


varungandhi-src (approved):

varungandhi-src:
We should update the `dev/bump-version` script to run prettier too, since the default ChangeLog formatting will fail a prettier check.

olafurpg:
I updated `bump-version` to run prettier

PR #97, opened 2022-04-28
Render documentation and relationships in the snapshot tests

olafurpg:
Previously, the snapshot tests only displayed occurrence roles and symbols. Now, the additionally display the `SymbolInformation.{documentation,relationships}` fields for definition occurrences.

### Test plan
See updated snapshot tests.

### DIFF
--- a/snapshots/output/react/src/LoaderInput.tsx
+++ b/snapshots/output/react/src/LoaderInput.tsx
@@ -9,11 +9,9 @@
     loading: boolean
 //  ^^^^^^^ definition react-example 1.0.0 src/`LoaderInput.tsx`/Props#loading.
 //  documentation ```ts\nboolean\n```
-//  documentation 
     children: React.ReactNode
 //  ^^^^^^^^ definition react-example 1.0.0 src/`LoaderInput.tsx`/Props#children.
 //  documentation ```ts\nReactNode\n```
-//  documentation 
 //            ^^^^^ reference @types/react 17.0.0 `index.d.ts`/React/
 //                  ^^^^^^^^^ reference @types/react 17.0.0 `index.d.ts`/React/ReactNode#
   }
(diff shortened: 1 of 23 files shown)


src/SnapshotTesting.ts
```diff
@@ -0,0 +1,88 @@
+import { Input } from './Input'
+import * as lsif from './lsif'
+import { Range } from './Range'
+
+const lsiftyped = lsif.lib.codeintel.lsiftyped
+
+export function formatSnapshot(
```
olafurpg, line 7: I took the liberty to extract this functionality into a separate file so that it's easier to reuse between projects.

src/FileIndexer.ts
```diff
@@ -141,16 +141,17 @@ export class FileIndexer {
     sym: ts.Symbol,
     symbol: LsifSymbol
   ): void {
+    const documentation = [
```
olafurpg, line 144: The diff in the snapshots was difficult to understand with this bugfix so I included it in the PR

src/SnapshotTesting.ts
olafurpg: I think it's fine to copy-paste this logic for now while the format is still evolving. The actual formatting logic is fairly simple to port between languages

snapshots/output/react/src/LoaderInput.tsx
```diff
@@ -4,16 +4,21 @@
   /** Takes loading prop, input component as child */
   interface Props {
 //          ^^^^^ definition react-example 1.0.0 src/`LoaderInput.tsx`/Props#
+//          documentation ```ts\nProps\n```
+//          documentation Takes loading prop, input component as child
```
varungandhi-src, line 8: Is repeating the `documentation` on multiple lines intentional? TJ's PR uses a separate character from the second line.

src/SnapshotTesting.ts
```diff
@@ -0,0 +1,88 @@
...
+      out.push('//')
+      out.push(' '.repeat(range.start.character))
+      const length = range.end.character - range.start.character
+      if (length < 0) {
+        throw new Error(input.format(range, 'negative length occurrence!'))
+      }
+      out.push('^'.repeat(length))
+      out.push(' ')
+      const isDefinition =
+        (occurrence.symbol_roles & lsiftyped.SymbolRole.Definition) > 0
```
varungandhi-src, line 45: Maybe `!!` or `!== 0` would be more idiomatic here rather than `> 0`?

varungandhi-src (approved):

snapshots/output/react/src/LoaderInput.tsx
olafurpg: The `documentation` part is repeated for separate documentation elements. The formatting of multiline docstrings is different, here newlines are rendered as an escaped `\n` character
CleanShot 2022-04-28 at 18 02 33

src/SnapshotTesting.ts
olafurpg: I'm not sure what's most idiomatic, this is how I would have written it in Java :/

PR #99, opened 2022-04-28
Add GitHub Action to index lsif-typescript's code.

varungandhi-src:
Based on the workflow in sourcegraph/sourcegraph repo.

### Test plan

n/a

### DIFF
--- a/.github/workflows/lsif-typescript.yml
+++ b/.github/workflows/lsif-typescript.yml
@@ -16,10 +16,8 @@ jobs:
         run: |
           curl -L https://sourcegraph.com/.api/src-cli/src_linux_amd64 -o /usr/local/bin/src
           chmod +x /usr/local/bin/src
-      - name: Install lsif-typescript
-        run: yarn run build
       - name: Run lsif-typescript
-        run: node ./dist/src/main.js index
+        run: ./node_modules/.bin/ts-node src/main.ts index --no-progress-bar
       - name: Upload index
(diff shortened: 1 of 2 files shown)


.github/workflows/lsif-typescript.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install npm dependencies
+        run: yarn --ignore-engines --ignore-scripts
+      - name: Install src-cli
+        run: |
+          curl -L https://sourcegraph.com/.api/src-cli/src_linux_amd64 -o /usr/local/bin/src
+          chmod +x /usr/local/bin/src
+      - name: Install lsif-typescript
+        run: yarn run build
```
olafurpg: We can run the TS files directly with ts-node. 
```suggestion
        run: ./node_modules/.bin/ts-node src/main.ts index --no-progress-bar
```
We may want to tweak the default value of `--no-progress-bar` so that it's automatically configured in CI environments. You can typically achieve this by either detecting the env variable `CI=true` or by testing for the presence of tty.

olafurpg (changes requested):

.github/workflows/lsif-typescript.yml
olafurpg: The benefit of `ts-node` is that it prints out stack traces from the original TS files instead of the compiled JS output.

varungandhi-src: Updated.

olafurpg (approved):

PR #100, opened 2022-04-28
Fix variable case: inferTsconfig -> inferTsConfig.

varungandhi-src:
### Test plan

Ran existing tests.

### DIFF
--- a/src/main.test.ts
+++ b/src/main.test.ts
@@ -44,17 +44,17 @@ for (const snapshotDirectory of snapshotDirectories) {
       fs.readFileSync(packageJsonPath).toString()
     ) as PackageJson
     const tsconfigJsonPath = path.join(inputRoot, 'tsconfig.json')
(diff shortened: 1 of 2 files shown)


olafurpg (changes requested):
This silently breaks the parsing of the `--infer-tsconfig` flag, which requires the field to be named `inferTsconfig`.

```
❯ ts-node ~/dev/sourcegraph/lsif-typescript/src/main.ts index --infer-tsconfig
- /Users/olafurpg/dev/sourcegraph-codeintel-showcase/lodash (missing tsconfig.json)
```

I think `--infer-tsconfig` is easier to read than `--infer-ts-config` because the config file is named `tsconfig.json`, not `ts-config.json`. I'll open a PR adding cli parsing tests for the most important use-cases.

olafurpg:
Superseded by #102

PR #102, opened 2022-04-29
Add tests for command-line parser

olafurpg:
Previously, we didn't have any tests for how we parsed command-line
flags. This was problematic because the tests didn't catch breaking
changes in seemingly innocent pull requests like https://github.com/sourcegraph/lsif-typescript/pull/100
that rename a variable. Now, we have tests for the command-line parser
so that we hopefully catch more unintentional regressions in the future.

### Test plan

See new `CommandLineOptions.test.ts` tests.


src/inferTsconfig.ts
```diff
@@ -19,7 +19,7 @@ export const noJsConfig = '{}'
  * If the directory contains at least one `*.{ts,tsx}` file then the config will be empty (`{}`).
  * If the directory doesn't contains one `*.{ts,tsx}` file then the config will
  */
-export function inferTsConfig(projectPath: string): string {
+export function inferTsconfig(projectPath: string): string {
```
olafurpg, line 22: I apologize for mixing this unrelated diff into the PR. It's quite painful to correct the capitalization of files on macOS and I didn't feel like doing it again to have a cleaner PR

src/CommandLineOptions.ts
```diff
@@ -0,0 +1,51 @@
...
+/** Configuration options to index a single TypeScript project. */
+export interface ProjectOptions extends MultiProjectOptions {
+  projectRoot: string
+  projectDisplayName: string
+  writeIndex: (index: lsif.lib.codeintel.lsiftyped.Index) => void
+}
+
+export function commanderCommand(
+  indexAction: (projects: string[], otpions: MultiProjectOptions) => void
+): Command {
```
varungandhi-src:
```suggestion
export function mainCommand(
  indexAction: (projects: string[], options: MultiProjectOptions) => void
): Command {
```

Maybe name the function to be something more generic/describing its use, rather than the implementation? The fact that we're using the `commander` library is somewhat incidental.

src/CommandLineOptions.test.ts
```diff
@@ -0,0 +1,36 @@
+import { test } from 'uvu'
+import * as assert from 'uvu/assert'
+
+import { commanderCommand, MultiProjectOptions } from './CommandLineOptions'
+
+function checkIndexParser(
+  args: string[],
+  expectedOptions: Partial<MultiProjectOptions>,
```
varungandhi-src, line 8: Oh wow, I didn't know this was a thing you could do so easily in TypeScript. (The `Partial<>` type)

src/CommandLineOptions.ts
```diff
@@ -0,0 +1,51 @@
...
+    .option('--cwd', 'the working directory', process.cwd())
+    .option('--yarn-workspaces', 'whether to index all yarn workspaces', false)
+    .option(
+      '--infer-tsconfig',
+      "whether to infer the tsconfig.json file, if it's missing",
+      false
+    )
+    .option('--output', 'path to the output file', 'dump.lsif-typed')
+    .option('--no-progress-bar', 'whether to disable the progress bar')
+    .argument('[projects...]')
```
varungandhi-src: Now I'm wondering... is there a difference between:

```
lsif-typescript index --cwd /some/path
vs
lsif-typescript index /some/path
```

?

src/CommandLineOptions.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export function commanderCommand(
+  indexAction: (projects: string[], otpions: MultiProjectOptions) => void
+): Command {
+  const command = new Command()
+  command
+    .name('lsif-typescript')
+    .version(packageJson.version)
+    .description('LSIF indexer for TypeScript and JavaScript')
```
varungandhi-src: Maybe add a line like:

```
For usage examples, see https://github.com/sourcegraph/lsif-typescript/blob/main/README.md
```

varungandhi-src (approved):

src/CommandLineOptions.ts
olafurpg: Great point, it was a bad name, renamed to `mainCommand`.

src/CommandLineOptions.ts
olafurpg: It's good you asked because it surfaced a bug in how we parse the `--cwd` flag. I fixed the bug and added a cli test for `--cwd`. There is a difference between those two:

- `index --cwd /path` generates `/path/dump.lsif-typed`
- `index /path` generates `$PWD/dump.lsif-typed`

src/CommandLineOptions.ts
olafurpg: Good idea, done.

src/CommandLineOptions.test.ts
olafurpg: It's very handy for cases like this! TypeScript includes several utility types https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype but you can go a bit overboard with them so use with care.

PR #135, opened 2022-05-20
Emit index.scip instead of dump.lsif-typed

olafurpg:
### Test plan

Manually run `scip-typescript index` and see it produce an `index.scip` file instead of `dump.lsif-typed`


olafurpg:
Merging since we need a release of this change before we can change from lsif-typescript to scip-typescript for auto-indexing.

PR #136, opened 2022-05-24
Implement "Find implementations" for classes, properties, and methods

olafurpg:
Previously, scip-typescript did not support "Find implementations". Now,
it supports this functionality for a few commonly used patterns where
inheritance is used. This PR does not implement all locations where
inheritance is used, but we can add that functionality later.

### Test plan

See the newly added snapshot tests. I also manually ran the indexer against sourcegraph/sourcegraph to verify that these changes don't introduce new unexpected crashes when running against a larger codebase. This change adds maybe ~10% overhead based on very rough observations but I feel like we can optimize some parts if necessary (for example, cache the logic that walks through parents).

### DIFF
--- a/snapshots/output/syntax/src/inheritance.ts
+++ b/snapshots/output/syntax/src/inheritance.ts
@@ -1,3 +1,6 @@
+  import { Overloader } from './overload'
+//         ^^^^^^^^^^ reference syntax 1.0.0 src/`overload.d.ts`/Overloader#
+  
   export interface Superinterface {
 //                 ^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`inheritance.ts`/Superinterface#
 //                 documentation ```ts\ninterface Superinterface\n```
@@ -28,6 +31,12 @@
 //                      documentation ```ts\nclass IntermediateSuperclass\n```
 //                      relationship implementation scip-typescript npm syntax 1.0.0 src/`inheritance.ts`/Superclass#
 //                                                     ^^^^^^^^^^ reference syntax 1.0.0 src/`inheritance.ts`/Superclass#
+    public override overrideMethod(): string {
+//                  ^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`inheritance.ts`/IntermediateSuperclass#overrideMethod().
+//                  documentation ```ts\n(method) overrideMethod(): string\n```
+//                  relationship implementation reference scip-typescript npm syntax 1.0.0 src/`inheritance.ts`/Superclass#overrideMethod().
+      return 'this will get overridden'
+    }
     public abstract intermediateOverrideMethod(): string
 //                  ^^^^^^^^^^^^^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`inheritance.ts`/IntermediateSuperclass#intermediateOverrideMethod().
 //                  documentation ```ts\n(method) intermediateOverrideMethod(): string\n```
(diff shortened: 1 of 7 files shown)


snapshots/output/syntax/src/inheritance.ts
```diff
@@ -0,0 +1,90 @@
...
+//              ^^^^^ reference typescript 4.6.2 lib/`lib.es5.d.ts`/Error#
+//              ^^^^^ reference typescript 4.6.2 lib/`lib.es5.d.ts`/Error.
+    }
+    public interfaceMethod(): string {
+//         ^^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`inheritance.ts`/Subclass#interfaceMethod().
+//         documentation ```ts\n(method) interfaceMethod(): string\n```
+//         relationship implementation scip-typescript npm syntax 1.0.0 src/`inheritance.ts`/Superinterface#interfaceMethod().
+      throw new Error('Method not implemented.')
+//              ^^^^^ reference typescript 4.6.2 lib/`lib.es5.d.ts`/Error#
+//              ^^^^^ reference typescript 4.6.2 lib/`lib.es5.d.ts`/Error.
```
varungandhi-src, line 96: Unrelated to this PR, I think, but it seems a little off that there are 2 references being emitted here instead of one.

src/FileIndexer.ts
```diff
@@ -148,11 +150,55 @@ export class FileIndexer {
...
-      new lsiftyped.SymbolInformation({ symbol: symbol.value, documentation })
+      new lsiftyped.SymbolInformation({
+        symbol: symbol.value,
+        documentation,
+        relationships: this.scipRelationships(declaration, symbol),
+      })
     )
   }
 
+  private scipRelationships(
```
varungandhi-src:
```suggestion
  private relationships(
```

nit: It seems a little weird to have scip in the method name; other methods don't seem to have this.

src/FileIndexer.ts
```diff
@@ -148,11 +150,55 @@ export class FileIndexer {
...
+      ts.isPropertyAssignment(declaration) ||
+      ts.isPropertyDeclaration(declaration)
+    ) {
+      this.forEachParent(declaration.parent, parent => {
+        for (const member of parent.members) {
+          if (declaration.name.getText() === member.name?.getText()) {
+            pushImplementation(this.lsifSymbol(member))
+          }
+        }
+      })
```
varungandhi-src: Is it possible for one method or property to:
1. Implement multiple members from the same parent -- If not, we should add a `return` after the `pushImplementation`.
2. Implement members from different parents -- If not, the lambda should have a return type of `"continue" | "break"` we should break the `forEachParent` loop by returning a `"break"` value from the lambda.

Small bit but maybe hoist out `declaration.name.getText()` outside the `forEachParent`?

src/FileIndexer.ts
```diff
@@ -397,6 +443,81 @@ export class FileIndexer {
...
+  // that node. The callback is invoked on the `node` parameter itself if it's
+  // class-like or an interface.
+  private forEachParent(
+    node: ts.Node,
+    onParent: (
+      parent: ts.ClassLikeDeclaration | ts.InterfaceDeclaration
+    ) => void
+  ): void {
+    const isVisited = new Set<ts.Node>()
+    const loop = (declaration: ts.Node): void => {
```
varungandhi-src, line 471: I don't know if it is possible to have cyclic inheritance chains somehow, but maybe we should add a fuel limit here to avoid a stack overflow if someone manages to write that kind of code?

If we run out of fuel, we can log a diagnostic message.

src/FileIndexer.ts
```diff
@@ -397,6 +443,81 @@ export class FileIndexer {
...
+        for (const heritageClause of declaration?.heritageClauses || []) {
+          for (const tpe of heritageClause.types) {
+            const parentSymbol = this.getTSSymbolAtLocation(tpe.expression)
+            if (parentSymbol) {
+              for (const parentDecl of parentSymbol.declarations || []) {
+                loop(parentDecl)
+              }
+            }
+          }
+        }
```
varungandhi-src: Could you add a TODO here that we should handle the parents of interfaces too? (And briefly mention which other other cases are missing.)

I'm guessing this is what you are referring to in the PR description

> This PR does not implement all locations where inheritance is used, but we can add that functionality later.

src/FileIndexer.ts
olafurpg: 1. I think it's possible for a single method to implement multiple overloaded abstract methods from type definitions. For example, it's common for type definitions to provide overloads with different string literals. The JS runtime has no concept of overloading so if you have a concrete implementation of a method named `x` then it must implement all abstract method definitions in the parent types with the matching name `x`. That's my understanding at least.
2. Can you elaborate? On the JVM you can specifically override a method from a specific supertype but I don't think the JS runtime allows this.
3. Done.

src/FileIndexer.ts
olafurpg: Done. This was actually named `relationships` in my initial draft but I renamed to `scipRelationships` for consistency with `lsifSymbol` (which should be renamed into `scipSymbol` but I wanted to minimize unrelated diffs). Either way, I'm happy with `relationships`.

src/FileIndexer.ts
olafurpg: Cycles are already guarded with the `isVisited` set, which also handles diamond dependencies

src/FileIndexer.ts
olafurpg: I added an `|| ts.isInterfaceDeclaration(declaration)` so that we recursively traverse both class and interface parents. I don't have any examples at the top of my mind that are not handled besides object literals that implement interfaces need an occurrence (local symbol) so that "find implementation" on the interface symbol itself goes to the object literal. I opened an issue to track this case https://github.com/sourcegraph/scip-typescript/issues/138

src/FileIndexer.ts
varungandhi-src: Oh right, we are marking it as visited right away, not at the end. 👍

src/FileIndexer.ts
varungandhi-src: 1. Interesting! (Non-blocking suggestion) Can you add a test case for that situation?
2. I mean that if we found a match in one interface (say), does it make sense to keep traversing and potentially find other matches in other ancestor interfaces (which is what the code will do right now IIUC)? If not, we should stop the traversal in `forEachParent` (i.e. break the loop) by returning a result from the lambda indicating that. Otherwise, we're wasting time looking at more ancestors even though we already found a match. If it is meaningful to have multiple matches, it would be good to have a test case demonstrating that.

src/FileIndexer.ts
```diff
@@ -397,6 +444,84 @@ export class FileIndexer {
     }
     return node.getText() + ': ' + type()
   }
+
+  // Invokes the `onParent` callback for all "parents" of the provided node,
+  // where "parent" is loosely defined as the superclass or superinterface of
+  // that node. The callback is invoked on the `node` parameter itself if it's
+  // class-like or an interface.
+  private forEachParent(
```
varungandhi-src: nit: Should this be called `forEachAncestor` instead? (If so, please rename the argument + update the doc comment above) IIUC that would be more accurate; using "parent" seems confusing because that seems to imply that the traversal will only go 1 level deep.

snapshots/output/syntax/src/inheritance.ts
olafurpg: VS Code shows three definition symbols here where we only show two, we're also missing the constructor

CleanShot 2022-05-25 at 08 25 55@2x

src/FileIndexer.ts
olafurpg: 1. Done. Given that we emit the same SCIP symbol for overloaded TS methods then it doesn't make a practical difference. But at least we have a test case for this situation now.
2. I expect "Find implementations" to display all implementations, including transitive ones. For example, the implementations of `Animal#sound().` are both `Mammal#sound().` and `Animal#sound().`

```ts
interface Animal { sound: string }
class Mammal implements Animal { sound = "mammal" }
class Dog extends Mammal { sound = "woof" }
```

PR #145, opened 2022-06-02
Install src via npm instead of curl command

olafurpg:
### Test plan

See the updated CI job go green.


tjdevries (approved):
this looks so much better, thank you

PR #168, opened 2022-08-26
Add support for Yarn v3

olafurpg:
### Test plan

I ran `npx ts-node ~/dev/sourcegraph/scip-typescript/src/main.ts index --yarn-berry-workspaces` locally and it successfully indexed the Sourcegraph repo from this PR here https://github.com/sourcegraph/sourcegraph/pull/39728

### DIFF
--- a/src/CommandLineOptions.ts
+++ b/src/CommandLineOptions.ts
@@ -9,6 +9,7 @@ export interface MultiProjectOptions {
   inferTsconfig: boolean
   progressBar: boolean
   yarnWorkspaces: boolean
+  yarnBerryWorkspaces: boolean
   cwd: string
   output: string
   indexedProjects: Set<string>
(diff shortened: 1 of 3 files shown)


src/CommandLineOptions.ts
```diff
@@ -35,6 +36,7 @@ export function mainCommand(
     .command('index')
     .option('--cwd <path>', 'the working directory', process.cwd())
     .option('--yarn-workspaces', 'whether to index all yarn workspaces', false)
+    .option('--yarn-berry-workspaces', 'whether to index all yarn v3 workspaces', false)
```
olafurpg: @valerybugakov would it make sense to name this flag `--yarn-v3-workspaces`?

olafurpg (commented):
I think we can automatically detect Yarn v3 in the future so that users don't need to use different flags based on what Yarn version they have, but for now I think it's OK to implement v3 support through a separate flag.

src/CommandLineOptions.ts
valerybugakov: @olafurpg, my rationale behind `berry` naming here is that this API was changed in the v2 and probably will be stable in v4 — it came with the `berry` rewrite of `yarn`. Open to other ideas, but tieing it to the specific `yarn` version can be misleading.

olafurpg: Fine by me

valerybugakov (approved):
Thank you for your quick response, Olaf! 💜

PR #176, opened 2022-10-11
Update workspaces list max buffer to 5 MB

pastelsky:
On very large workspaces, the output of `yarn workspaces info` might not fit in the default buffer of 200 KB. This increases it to 5MB — which shouldn't have any inadvertent side effects.

### Test plan

Manually tested

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -206,6 +206,7 @@ function listYarnBerryWorkspaces(directory: string): string[] {
     .execSync('yarn workspaces list --json', {
       cwd: directory,
       encoding: 'utf-8',
+      maxBuffer: 1024 * 1024, // 1MB
     })
     .split('\n')
   for (const line of lines) {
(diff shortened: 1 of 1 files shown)


varungandhi-src:
I don't see any problems with this, but curious about where you got the number of 200KB from? I see a 1MiB limit here with Node v18.10: https://nodejs.org/api/child_process.html

src/main.ts
```diff
@@ -206,6 +206,7 @@ function listYarnBerryWorkspaces(directory: string): string[] {
     .execSync('yarn workspaces list --json', {
       cwd: directory,
       encoding: 'utf-8',
+      maxBuffer: 1024 * 1024 * 5, // 5MB
```
varungandhi-src, line 209:
```suggestion
      maxBuffer: 1024 * 1024 * 5,
```

varungandhi-src (approved):

pastelsky:
Perhaps the limit changed? 
I was using this as reference-
https://github.com/nodejs/node/issues/4236

PR #181, opened 2022-10-12
fix: Fix --yarn-workspace with Yarn 3.

### LINKED ISSUE #180: scip-typescript fails to index Yarn workspaces with Yarn 3
Using the `--yarn-workspaces` flag triggers an invocation to ` yarn --silent --json workspaces info`. However, `workspaces info` doesn't seem to be available with Yarn 3.2.3. It is available with Yarn 1.22.19.

varungandhi-src:
Fixes #180.

### Test plan

Manually tested against the Sourcegraph monorepo with Yarn 3 which seems to have been made the default around Sept 1 (http://github.com/sourcegraph/sourcegraph/pull/39728). The last scip-typescript upload for sg/sg was on Sept 2 (https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/code-graph/uploads?query=scip-typescript), so it seems like this job has been broken since then.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -245,9 +245,9 @@ function listYarnWorkspaces(
   }
   if (yarnVersion === 'tryYarn1') {
     try {
-      yarn1WorkspaceInfo()
-    } catch {
       yarn2PlusWorkspaceInfo()
+    } catch {
+      yarn1WorkspaceInfo()
     }
   } else {
     yarn2PlusWorkspaceInfo()
--- a/src/main.ts
+++ b/src/main.ts
@@ -224,26 +224,42 @@ function listYarnBerryWorkspaces(directory: string): string[] {
 }
 
 function listYarnWorkspaces(directory: string): string[] {
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-  const json = JSON.parse(
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access
-    JSON.parse(
-      child_process.execSync('yarn --silent --json workspaces info', {
-        cwd: directory,
-        encoding: 'utf-8',
-        maxBuffer: 1024 * 1024 * 5, // 5MB
-      })
-    ).data
-  )
-
+  const runYarn = (cmd: string): string =>
+    child_process.execSync(cmd, {
+      cwd: directory,
+      encoding: 'utf-8',
+      maxBuffer: 1024 * 1024 * 5, // 5MB
+    })
   const result: string[] = []
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-  for (const key of Object.keys(json)) {
-    const location = 'location'
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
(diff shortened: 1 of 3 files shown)


olafurpg (commented):
What's the relationship with this PR here https://github.com/sourcegraph/scip-typescript/pull/168? We already support Yarn v3 with `--yarn-berry-workspaces`

varungandhi-src:
Hmm, I didn't know what Yarn Berry was. I guess it is a little weird to have two flags instead of one...

varungandhi-src:
Looks like I just reimplemented that PR but in a different function. 🤦🏽 

I thought Yarn Berry was some different fork of Yarn or something that we added special support for.

olafurpg:
cc/ @valerybugakov any opinions about having a separate flag for Yarn v3?

valerybugakov:
@olafurpg, I don't see a reason for having a separate flag for Yarn v3 if we already have `--yarn-berry-workspaces`. There's the next major `yarn` version on the horizon, so it makes sense to bind the flag to the package-manager workspaces protocol, not the package-manager version.

olafurpg:
@valerybugakov I meant, should we aim to consolidate `--yarn-workspaces` and `--yarn-berry-workspaces`? Varun implemented this PR without realizing we already support v3 because the `--yarn-berry-workspaces` name is not self-explanatory for somebody who is not intimately familiar with Yarn development

valerybugakov:
> should we aim to consolidate `--yarn-workspaces` and `--yarn-berry-workspaces`?

I see. Then yes, hiding this detail from the user would be nice 👍

varungandhi-src:
OK, looks like we're in agreement then. I've removed the code duplication between the newly added implementation and the previously added one.

src/main.ts
```diff
@@ -200,51 +200,57 @@ function defaultCompilerOptions(configFileName?: string): ts.CompilerOptions {
...
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-  for (const key of Object.keys(json)) {
-    const location = 'location'
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-    if (json[key][location] !== undefined) {
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access
-      result.push(path.join(directory, json[key][location]))
+  if (yarnVersion === 'tryYarn1') {
+    try {
+      yarn1WorkspaceInfo()
```
olafurpg: Do we flip the order here?

olafurpg (approved):
LGTM 👍🏻

src/main.ts
```diff
@@ -200,51 +200,57 @@ function defaultCompilerOptions(configFileName?: string): ts.CompilerOptions {
...
+    } catch {
+      yarn2PlusWorkspaceInfo()
```
varungandhi-src:
```suggestion
      yarn2PlusWorkspaceInfo()
    } catch {
      yarn1WorkspaceInfo()
```

olafurpg (commented):
Also worth updating the readme https://github.com/sourcegraph/scip-typescript#index-a-typescript-project-using-yarn-workspaces

Issue #175, opened 2022-10-11
Process packages in parallel


pastelsky:
`scip-typescript` currently can take several minutes to index yarn workspaces. It does not seem to saturate CPU enough or use all cores. Ideally there should be an option to do indexing for multiple packages concurrently to reduce this time.

This would be useful considering slow indexing reduces the benefit of type indexing in the first place on a repository with high frequency commits as a new commits can invalidate the ongoing index.

varungandhi-src:
Could you share a rough number for how many lines of code / second are you seeing indexing at and what your target number is based on workspace size and peak commit velocity? IIRC, one of the problems here was that we didn't have easy access to the dependency graph between packages in a workspace, which meant the determining the order of indexing was non-trivial, and it would be wasteful to do the naive thing of indexing in parallel without understanding the dependency graph as that would likely lead to a bunch of duplicated work.

Even if we did things in dependency order, it might not be fast enough if your build graph is not wide enough.

pastelsky:
I'm currently trying to index a large monorepo with over 600+ packages, so I'm a sure topologically sorted graph would be wide enough to saturate CPUs.

 > Could you share a rough number for how many lines of code / second are you seeing indexing at

It roughly takes me 20 mins to index on a repository with ~2M SLOC of typescript on an M1 Macbook Pro.

> what your target number is based on workspace size and peak commit velocity.

Master for us gets updated roughly once every 15 minutes (but this could be higher going forward). Ideally we'd want to be able to index + compress + upload + process well within this time. But given that sourcegrpah does not invalidate the entire index when a new commit is pushed (IIRC from docs), it might not be too bad if the time exceeds that by a bit. Still, the faster, the better.

pastelsky:
> it would be wasteful to do the naive thing of indexing in parallel without understanding the dependency graph as that would likely lead to a bunch of duplicated work.

I think even with some duplicated work, having an option to use more cores might be several times faster on a multi-core machine, and that could be just enough for some repositories to exceed their master update rate. Perhaps this can be something that can be made more efficient over time rather than not providing an option at all ?

I'm happy to test out any early changes and give feedback on it from our repositories if it helps. Very interested in making this scale for us at the moment (and should hopefully help other enterprise users of sourcegraph too)

varungandhi-src:
> But given that sourcegrpah does not invalidate the entire index when a new commit is pushed

Right, code navigation will fall back to the nearest index if one can't be found for the latest commit.

[...]

PR #182, opened 2022-10-17
Optimization: cache results between TypeScript projects

olafurpg:
Towards https://github.com/sourcegraph/scip-typescript/issues/175

Previously, scip-typescript didn't cache anything at all between TypeScript projects. This commit implements an optimization so that we now cache the results of loading source files and parsing options. Benchmarks against the sourcegraph/sourcegraph repo indicate this optimization consistently speeds up the `index` command in all three multi-project repositories that I tested it with.

- sourcegraph/sourcegraph: ~30% from ~100s to ~70s
- nextautjs/next-auth: ~40% from 6.5s to 3.9
- xtermjs/xterm.js: ~45% from 7.3s to 4.1s

For every repo, I additionally validated that the resulting index.scip has identical checksum before and after applying this optimization. Given these promising results, this new optimization is enabled by default, but can be disabled with the option `--no-global-cache`.

### Test plan

Manually tested by running `scip-typescript index tsconfig.all.json` in the sourcegraph/sourcegraph repository. To benchmark the difference for this PR:

- Checkout the code
- Run `yarn tsc -b`
- Go to the directory of your project
- Run `node PATH_TO_SCIP_TYPESCRIPT/dist/src/main.js`
- Copy the "optimized" index.scip with `cp index.scip index-withcache.scip`
- Run `node PATH_TO_SCIP_TYPESCRIPT/dist/src/main.js --no-global-caches`
- Validate the checksum is identical from the optimized output `shasum -a 256 *.scip`


src/ProjectIndexer.ts
```diff
@@ -1,17 +1,53 @@
...
+    return host
+  }
+  const hostCopy = { ...host }
+  host.getParsedCommandLine = (fileName: string) => {
+    if (!hostCopy.getParsedCommandLine) {
+      return undefined
+    }
+    if (cache.parsedCommandLines.has(fileName)) {
+      return cache.parsedCommandLines.get(fileName)
+    }
```
varungandhi-src: I don't fully understand all the cases here; is it possible that the cache has a hit for this even though `hostCopy.getParsedCommandLine` is missing? (when is `getParsedCommandLine` missing?)

src/ProjectIndexer.ts
```diff
@@ -1,17 +1,53 @@
...
+  host.getSourceFile = (fileName, languageVersion) => {
+    const fromCache = cache.sources.get(fileName)
+    if (fromCache !== undefined) {
+      return fromCache
+    }
+    const result = hostCopy.getSourceFile(fileName, languageVersion)
+    if (result !== undefined) {
+      cache.sources.set(fileName, result)
+    }
+    return result
```
varungandhi-src: suggestion: Maybe add a comment here or near the cache definition which says "It is safe use the file name as a key here because we assume that a source file only ever belongs to a single project, and is only ever type-checked with a single language version".

varungandhi-src:
Very nice! ~How did you recognize that this approach of caching the SourceFile values would help speed up type-checking?~ (Nvm, saw that you wrote this in the issue.)

src/CommandLineOptions.test.ts
```diff
@@ -24,7 +24,6 @@ function checkIndexParser(
 
 // defaults
 checkIndexParser([], {
-  progressBar: true,
```
varungandhi-src, line 27: Why did this get removed? Shouldn't there be an extra `globalCaches: true` here instead?

varungandhi-src (approved):

src/CommandLineOptions.test.ts
olafurpg: I forgot to mention it in the description but this PR also disables the fancy progress bar by default. I didn't measure any performance difference with it enabled/disabled, but I found it distracting while testing projects. scip-typescript runs >99% of the time in CI where you should disable it by default anyways and I'm guessing most people are not doing it.

src/ProjectIndexer.ts
olafurpg: I may be misunderstanding your comment, but we guard against `hostCopy.getParsedCommandLine` at the top, so even if the cache has a hit we return undefined when `hostCopy.getParsedCommandLine` is undefined.

src/ProjectIndexer.ts
olafurpg: Good call! I added an additional check to additionally only cache when `languageVersion` is unchanged (excluding a `setExternalModuleIndicator` field that is a function that is never the same. I also added a comment explaining in more details. I also discovered that `getSourceFile` has two optional parameters that we're now properly forwarding.

olafurpg (commented):
Thank you for the review! 🙏🏻

Issue #183, opened 2022-10-18
Redundant occurrences for wildcard import statement


olafurpg:
While snapshotting index.scip for the scip-typescript repo itself I saw this output
```ts
  import * as ts from 'typescript'
//^ definition scip-typescript npm @sourcegraph/scip-typescript 0.2.11 src/FileIndexer.ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.8.4 lib/typescript.d.ts/ts/
[...]

PR #184, opened 2022-10-18
Emit occurrences for string literals

olafurpg:
Previously, scip-typescript only emitted occurrences for identifiers. Now, we additionally emit occurrences for string literals so that features like "Go to definition" work with `import` statements and other locations where string literals reference actual symbols.

### Test plan

See the updated snapshot tests. I additionally manually verified that it works as expected on Sourcegraph https://sourcegraph.com/github.com/sourcegraph/scip-typescript@324cb332db75b3082fdbccc45a1a8e166d546b28/-/blob/src/FileIndexer.ts?L21:28&popover=pinned

CleanShot 2022-10-18 at 11 39 31@2x

### DIFF
--- a/snapshots/output/syntax/src/import.ts
+++ b/snapshots/output/syntax/src/import.ts
@@ -1,6 +1,6 @@
   import { Class } from './class'
 // definition syntax 1.0.0 src/`import.ts`/
-//documentation ```ts\nmodule "import"\n```
+//documentation ```ts\nmodule "import.ts"\n```
 //         ^^^^^ reference syntax 1.0.0 src/`class.ts`/Class#
 //                      ^^^^^^^^^ reference syntax 1.0.0 src/`class.ts`/
   import { Enum } from './enum'
--- a/snapshots/output/syntax/src/import.ts
+++ b/snapshots/output/syntax/src/import.ts
@@ -45,7 +45,7 @@
 //                ^^^^^^^^^^^^ reference syntax 1.0.0 src/`function.ts`/
 //                              ^^^^ reference typescript 4.8.4 lib/`lib.es5.d.ts`/Promise#then().
 //                                   ^ definition local 3
-//                                   documentation ```ts\n(parameter) c: typeof import("/Users/olafurpg/dev/sourcegraph/scip-typescript/snapshots/input/syntax/src/function")\n```
+//                                   documentation ```ts\n(parameter) c: typeof import("/src/function")\n```
 //                                        ^ reference local 3
 //                                          ^^^^^^^^^^^ reference syntax 1.0.0 src/`function.ts`/newFunction().
   }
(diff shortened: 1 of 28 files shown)


snapshots/output/syntax/src/import.ts
```diff
@@ -27,3 +33,20 @@
...
   }
   
+  export function dynamicImport(): Promise<void> {
+//                ^^^^^^^^^^^^^ definition syntax 1.0.0 src/`import.ts`/dynamicImport().
+//                documentation ```ts\nfunction dynamicImport(): Promise<void>\n```
+//                                 ^^^^^^^ reference typescript 4.8.4 lib/`lib.es5.d.ts`/Promise#
+//                                 ^^^^^^^ reference typescript 4.8.4 lib/`lib.es2015.iterable.d.ts`/Promise#
+//                                 ^^^^^^^ reference typescript 4.8.4 lib/`lib.es2015.promise.d.ts`/Promise.
+//                                 ^^^^^^^ reference typescript 4.8.4 lib/`lib.es2015.symbol.wellknown.d.ts`/Promise#
+//                                 ^^^^^^^ reference typescript 4.8.4 lib/`lib.es2018.promise.d.ts`/Promise#
```
varungandhi-src, line 43: Unrelated to this PR, but is it intentional that we have duplicate references for es2015 and es2018?

olafurpg: I opened https://github.com/sourcegraph/scip-typescript/issues/183 earlier today

src/FileIndexer.ts
```diff
@@ -25,22 +28,52 @@ export class FileIndexer {
...
+        symbol: symbol.value,
+        symbol_roles: lsiftyped.SymbolRole.Definition,
+      })
+    )
+    const moduleName =
+      this.sourceFile.moduleName ||
+      path
+        .basename(this.sourceFile.fileName)
+        .replace(/.tsx?$/, '')
+        .replace(/.jsx?/, '')
```
varungandhi-src: 1. Is it ever possible that we have files with the same base name but with different extensions? IIUC this happens with translation/conversion, but we will not index the generated files.
2. I think you need to add `.d.ts`, `.cjs` and `.mjs` extensions too.
3. Since the `.` is inside the regex, it needs escaping, right? Or is this some reason why it doesn't need to be escaped here?

I'm wondering, instead of dropping the extension here, can we get the filename and include the extension when emitting the occurrence for the import instead? Any reason to prefer dropping the extension instead of doing that?

olafurpg: This case should not happen since it's a fallback when `moduleName` is undefined, and I'm guessing it's always defined for `ts.SourceFile`

olafurpg: My preference for dropping the extension was to best-effort mirror the output of `moduleName`, even if it doesn't handle all cases.

varungandhi-src: Hmm, is the replacement even working? The occurrences seem to have the file extensions...

varungandhi-src (approved):

src/FileIndexer.ts
olafurpg: Turns out `moduleName` was in fact undefined almost everywhere! I updated the logic to keep the file extension so that we don't treat d.ts/mjs/cjs extensions differently

PR #185, opened 2022-10-18
Emit correct occurrences for destructured objects

olafurpg:
Previously, scip-typescript didn't emit appropriate occurrences for `property` in the destructuring pattern below:
```ts
interface Props { property: number }
const prop = { property: 42 }
const { property } = prop
//      ^^^^^^^^ before: was a local reference
//               now: local definition + global reference
```
The solution works similarly to how to handle shorthand property definitions.

### Test plan

See the updated snapshot tests.

### DIFF
--- a/snapshots/input/syntax/src/destructuring.ts
+++ b/snapshots/input/syntax/src/destructuring.ts
@@ -31,3 +31,7 @@ export function forLoopArrayDestructuring(): number {
   }
   return 1
 }
+
+export function parameterDestructuring({ a }: Props): number {
+  return a
+}
--- a/snapshots/input/syntax/src/destructuring.ts
+++ b/snapshots/input/syntax/src/destructuring.ts
@@ -1,0 +1,33 @@
+interface Props {
+  a: number
+}
+const props: Props = { a: 42 }
+
+export function objectDestructuring(): number[] {
+  const { a: b } = props
+  return [props].map(({ a }) => a + b)
+}
+
+export function arrayDestructuring(): number[] {
+  const [b] = [props]
+  return [[b]].map(([a]) => a.a)
+}
+
+export function nestedDestructuring(): number[] {
+  const [[b]] = [[props]]
+  return [[props]].map(([{ a }]) => a + b.a)
+}
+
+export function forLoopObjectDestructuring(): number {
+  for (const { a } of [props]) {
+    return a
+  }
+  return 1
+}
+
+export function forLoopArrayDestructuring(): number {
+  for (const [{ a }] of [[props]]) {
+    return a
+  }
+  return 1
+}
--- a/snapshots/input/syntax/src/structural-type.ts
+++ b/snapshots/input/syntax/src/structural-type.ts
@@ -1,0 +1,9 @@
+export function foo(): Promise<{ member: number }> {
+  return Promise.resolve({ member: 42 })
+}
+export function bar(): Promise<number> {
+  return foo().then(x => x.member)
+}
+export function bar2(): Promise<number> {
+  return foo().then(({ member }) => member)
+}
--- a/snapshots/input/syntax/src/structural-type.ts
+++ b/snapshots/input/syntax/src/structural-type.ts
@@ -7,3 +7,8 @@ export function bar(): Promise<number> {
 export function bar2(): Promise<number> {
   return foo().then(({ member }) => member)
 }
+
+type OptionsFlags<Type> = { [Property in keyof Type]: boolean }
+type FeatureFlags = { darkMode: () => void }
+export type FeatureOptions = OptionsFlags<FeatureFlags> // implicitly // type FeatureOptions = { // darkMode: boolean; // } const fo: FeatureOptions = { darkMode: true }; // ^ go to def
+export const fo: FeatureOptions = { darkMode: true }
--- a/snapshots/output/react/src/LoaderInput.tsx
+++ b/snapshots/output/react/src/LoaderInput.tsx
@@ -26,9 +26,14 @@
 //                                ^^^^^^^^^^^^^^^^^ reference @types/react 17.0.0 `index.d.ts`/React/FunctionComponent#
 //                                                  ^^^^^ reference react-example 1.0.0 src/`LoaderInput.tsx`/Props#
     loading,
-//  ^^^^^^^ reference local 3
+//  ^^^^^^^ definition local 3
+//  documentation ```ts\n(parameter) loading: boolean\n```
+//  ^^^^^^^ reference react-example 1.0.0 src/`LoaderInput.tsx`/Props#loading.
     children,
-//  ^^^^^^^^ reference local 4
+//  ^^^^^^^^ definition local 4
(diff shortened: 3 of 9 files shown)


snapshots/input/syntax/src/destructuring.ts
```diff
@@ -0,0 +1,33 @@
+interface Props {
+  a: number
+}
+const props: Props = { a: 42 }
+
+export function objectDestructuring(): number[] {
```
varungandhi-src, line 6: suggestion: Add a test for destructuring in function parameter lists (I think that's the only case which isn't covered here)

snapshots/input/syntax/src/structural-type.ts
```diff
@@ -0,0 +1,9 @@
+export function foo(): Promise<{ member: number }> {
+  return Promise.resolve({ member: 42 })
+}
```
varungandhi-src, line 3: What happens when there is a mapped type (https://www.typescriptlang.org/docs/handbook/2/mapped-types.html) involved as part of the definition? A mapped type can produce new object types, so the keys aren't visible in the source code, they're implicit as part of some logic.

E.g. if you have:

```
type OptionsFlags<Type> = {
  [Property in keyof Type]: boolean;
};

type FeatureFlags = {
  darkMode: () => void;
};
 
type FeatureOptions = OptionsFlags<FeatureFlags>;
// implicitly 
// type FeatureOptions = {
//    darkMode: boolean;
// }

const fo: FeatureOptions = { darkMode: true };
                          // ^ go to def
```

Will go-to-def on `darkMode` in the object literal take you to `FeatureOptions`? Will it take you to `FeatureFlags`? Nowhere?

snapshots/output/react/src/LoaderInput.tsx
```diff
@@ -26,9 +26,14 @@
...
-//  ^^^^^^^ reference local 3
+//  ^^^^^^^ definition local 3
+//  documentation ```ts\n(parameter) loading: boolean\n```
+//  ^^^^^^^ reference react-example 1.0.0 src/`LoaderInput.tsx`/Props#loading.
     children,
-//  ^^^^^^^^ reference local 4
+//  ^^^^^^^^ definition local 4
+//  documentation ```ts\n(parameter) children: ReactNode\n```
+//  ^^^^^^^^ reference react-example 1.0.0 src/`LoaderInput.tsx`/Props#children.
+//  ^^^^^^^^ reference local 7
```
varungandhi-src, line 36: Why is there an extra `reference local` here? The `loading` field doesn't have this.

snapshots/output/syntax/src/destructuring.ts
```diff
@@ -0,0 +1,97 @@
...
+//                                        ^ reference local 28
+//                                          ^ reference syntax 1.0.0 src/`destructuring.ts`/Props#a.
+  }
+  
+  export function forLoopObjectDestructuring(): number {
+//                ^^^^^^^^^^^^^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`destructuring.ts`/forLoopObjectDestructuring().
+//                documentation ```ts\nfunction forLoopObjectDestructuring(): number\n```
+    for (const { a } of [props]) {
+//               ^ definition local 41
+//               documentation ```ts\nvar a: number\n```
```
varungandhi-src, line 75: This being a `var` and not a `const`/`let` is a bug, right? Filed https://github.com/sourcegraph/scip-typescript/issues/186

snapshots/output/syntax/src/structural-type.ts
```diff
@@ -0,0 +1,56 @@
...
+//                                 documentation ```ts\n(property) member: number\n```
+    return Promise.resolve({ member: 42 })
+//         ^^^^^^^ reference typescript 4.8.4 lib/`lib.es5.d.ts`/Promise#
+//         ^^^^^^^ reference typescript 4.8.4 lib/`lib.es2015.iterable.d.ts`/Promise#
+//         ^^^^^^^ reference typescript 4.8.4 lib/`lib.es2015.promise.d.ts`/Promise.
+//         ^^^^^^^ reference typescript 4.8.4 lib/`lib.es2015.symbol.wellknown.d.ts`/Promise#
+//         ^^^^^^^ reference typescript 4.8.4 lib/`lib.es2018.promise.d.ts`/Promise#
+//                 ^^^^^^^ reference typescript 4.8.4 lib/`lib.es2015.promise.d.ts`/PromiseConstructor#resolve().
+//                 ^^^^^^^ reference typescript 4.8.4 lib/`lib.es2015.promise.d.ts`/PromiseConstructor#resolve().
+//                           ^^^^^^ definition syntax 1.0.0 src/`structural-type.ts`/member0:
```
varungandhi-src, line 21: Are these two symbols deliberately different/should they be the same or have a reference relationship?

src/FileIndexer.ts
```diff
@@ -124,12 +125,47 @@ export class FileIndexer {
...
+   * ```
+   */
+  private handleObjectBindingPattern(node: ts.Node, range: number[]): void {
+    if (
+      !ts.isIdentifier(node) ||
+      !ts.isBindingElement(node.parent) ||
+      !ts.isObjectBindingPattern(node.parent.parent)
+    ) {
+      return
+    }
```
varungandhi-src:
```suggestion
    const isBinding = 
          ts.isIdentifier(node) &&
          ts.isBindingElement(node.parent) &&
          ts.isObjectBindingPattern(node.parent.parent)
    if (!isBinding) {
      return
    }
```
nit: I think having a single `!` and assigning a name would make things more readable.

varungandhi-src (approved):
LGTM except for some of the questions I've left inline.

snapshots/input/syntax/src/destructuring.ts
olafurpg: Good idea. Done, and it works out of the box

snapshots/input/syntax/src/structural-type.ts
olafurpg: Nice one. It doesn't work. I added a snapshot test and opened an issue to follow up on this https://github.com/sourcegraph/scip-typescript/issues/187

snapshots/output/react/src/LoaderInput.tsx
olafurpg: Good catch, opened https://github.com/sourcegraph/scip-typescript/issues/188

snapshots/output/syntax/src/destructuring.ts
olafurpg: Looks like a bug indeed, thank you for opening that

snapshots/output/syntax/src/structural-type.ts
olafurpg: Go to definition doesn't with tsserver for this case. I opened https://github.com/sourcegraph/scip-typescript/issues/189 because I think we can make it work anyways, and it would be cool to do better than tsserver

src/FileIndexer.ts
olafurpg: Fair enough. Done

PR #191, opened 2022-10-25
Remove remnants from "LSIF Typed"

olafurpg:
The codebase had a lot of outdated references to "LSIF Typed", which has been renamed to SCIP. This commit replaces these outdated references to avoid confusion for people who are reading the codebase for the first time.

### Test plan

See the CI go green. This should be a pure refactoring, no functional difference.


olafurpg:
cc/ @mrnugget @valerybugakov who both asked about the usage of LSIF in scip-typescript

src/Descriptor.ts
```diff
@@ -1,9 +1,9 @@
...
+import * as scip from './scip'
 
-type Descriptor = lsif.lib.codeintel.lsiftyped.Descriptor
-const Descriptor = lsif.lib.codeintel.lsiftyped.Descriptor
-type Suffix = lsif.lib.codeintel.lsiftyped.Descriptor.Suffix
-const Suffix = lsif.lib.codeintel.lsiftyped.Descriptor.Suffix
+type Descriptor = scip.scip.Descriptor
+const Descriptor = scip.scip.Descriptor
+type Suffix = scip.scip.Descriptor.Suffix
+const Suffix = scip.scip.Descriptor.Suffix
```
varungandhi-src, line 6: nit: Can we somehow avoid this being `scip.scip`. It's not a big deal, but it looks a little gross lol.

olafurpg (commented):
We still copy-paste scip.ts from the sourcegraph/scip repo. It would be nice to publish the bindings to npm so that we can add a normal package.json dependency instead. We can do that later, however, this PR is already a big improvement.

src/Descriptor.ts
olafurpg: I agree it's not nice and I'd +1 if we can solve it somehow. It's not a regression from this PR though, it was gross before as well

olafurpg:
Any idea why the CI is not triggering?

src/FileIndexer.ts
```diff
@@ -14,27 +14,23 @@ import {
   typeParameterDescriptor,
 } from './Descriptor'
 import { Input } from './Input'
-import * as lsif from './lsif'
-import { LsifSymbol } from './LsifSymbol'
-import { lsiftyped } from './main'
 import { Packages } from './Packages'
 import { Range } from './Range'
+import * as scip from './scip'
+import { ScipSymbol } from './ScipSymbol'
```
varungandhi-src, line 20: nit: Following JS/TS conventions, I think this should be called SCIPSymbol, similar to HTML and URL in browser APIs.

varungandhi-src (approved):

src/FileIndexer.ts
olafurpg: I recall seeing in relation to `XMLHttpRequest` that the convention is that only three-letter acronym should remain uppercase. I really dislike SCIPSymbol, it's easier IMO to read ScipSymbol so I prefer to keep it

varungandhi-src:
Nothing seems to have changed in the CI config, so I'm as confused as you are. Some transient GitHub issue? Maybe try pushing a dummy commit?

olafurpg:
Empty commit seems to have done the trick

PR #192, opened 2022-10-27
Use unique declaration references

### LINKED ISSUE #152: Snapshot output seems noisy; are we emitting needlessly redundant references?
If you run `scip snapshot` on the index generated from running scip-typescript on itself, the output you get for FileIndexer.ts starts with:

```
  import * as ts from 'typescript'
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
//            ^^ reference scip-typescript npm typescript 4.6.2 lib/typescript.d.ts/ts/
```

Are we emitting one reference per use of `ts` in the file? 😅 This should probably get compressed away, but it'd be nice to avoid doing the work in the first place.

valerybugakov:
Fixes https://github.com/sourcegraph/scip-typescript/issues/152

The too many references issue was caused by the fact that `ts` has multiple definitions in the same file, and we used this information to create a reference for each `namespace ts` block.

Screenshot 2022-10-24 at 18 18 52

This PR removes redundant references by filtering out duplicated SCIP symbol values.
This PR also fixes SCIP symbol generation for namespace imports. Covered by the updated snapshot test.

### Test plan

See the updated `snapshots/input/syntax/src/namespace.d.ts` spanshot test.


package.json
```diff
@@ -11,8 +11,8 @@
     "tslint": "tslint -p tsconfig.json --format stylish",
     "eslint": "eslint --cache '**/*.ts?(x)'",
     "build": "node ./node_modules/typescript/bin/tsc -b .",
-    "test": "uvu -r ts-node/register --ignore dist",
-    "update-snapshots": "uvu -r ts-node/register --ignore dist --update-snapshots",
+    "test": "TS_NODE_TRANSPILE_ONLY=true uvu -r ts-node/register --ignore dist",
+    "update-snapshots": "TS_NODE_TRANSPILE_ONLY=true uvu -r ts-node/register --ignore dist --update-snapshots",
```
varungandhi-src, line 15: Will this flag change skip type-checking of the test code? Or is that fine because the type-checking happens during the `build` step?

varungandhi-src:
@valerybugakov I think you accidentally committed generated files. I've opened a PR to fix the `.gitignore`. https://github.com/sourcegraph/scip-typescript/pull/193

package.json
valerybugakov: > is that fine because the type-checking happens during the build step?

Correct. We can skip it here to speed up the command and allow type errors for local development.

valerybugakov:
@varungandhi-src, approved your PR 👍

src/FileIndexer.ts
```diff
@@ -608,6 +629,25 @@ export class FileIndexer {
 
     return this.checker.getTypeAtLocation(node)
   }
+
+  /**
+   * @param sym ts.Symbol with declarations to process.
+   * @returns an array of unique non-empty SCIP symbol values extracted
+   * from ts.Symbol declarations.
+   */
+  private uniqueDeclarationSymbolValues(sym?: ts.Symbol): string[] {
```
varungandhi-src, line 638: nit: I don't have a strong opinion on this, but I've at least been trying to have things which don't require access to state be free functions (there are some examples at the end of this file). Makes it easier to understand which code is "pure" vs which code potentially modifies some state inside the indexer.

src/FileIndexer.ts
```diff
@@ -305,7 +322,7 @@ export class FileIndexer {
     return undefined
   }
 
-  private scipSymbol(node: ts.Node): ScipSymbol {
+  private scipSymbol = (node: ts.Node): ScipSymbol => {
```
varungandhi-src, line 325: nit: Superfluous change?

src/FileIndexer.ts
```diff
@@ -65,7 +65,21 @@ export class FileIndexer {
...
+         *       ^ node.parent.name
+         *       ^^^^^ node.parent
+         */
+        const isDefinition = this.declarationName(node.parent) === node
+
+        if (isDefinition) {
+          return this.indexDefinition(node, sym)
+        }
+
+        this.indexReference(node, sym)
```
varungandhi-src: nit: Maybe `visitDefinition` and `visitReference`? We're using "index" as a verb for documents/projects, but for an AST traversal, "visit" seems more natural.

varungandhi-src (approved):

src/FileIndexer.ts
valerybugakov: @varungandhi-src, we call `this.scipSymbol(declaration)` in this method that uses the `FileIndexer` state.

src/FileIndexer.ts
valerybugakov: This change is required for a call in the `uniqueDeclarationSymbolValues` to bind the `scipSymbol` method to the `FileIndexer` instance.

src/FileIndexer.ts
varungandhi-src: Oh sorry, my bad. 🤦🏽

src/FileIndexer.ts
varungandhi-src: I'm confused...  other methods don't use this declaration style, and they can be called just fine... Is it because of the lambda in the `uniqueDeclarationSymbolValues` method?

olafurpg:
Please wait with merging until I've reviewed.

src/FileIndexer.ts
```diff
@@ -608,6 +629,25 @@ export class FileIndexer {
...
+      return []
+    }
+
+    return Array.from(
+      new Set(
+        sym.declarations
+          .map(declaration => this.scipSymbol(declaration).value)
+          .filter(Boolean)
+      )
+    )
```
varungandhi-src, line 649: Now I'm wondering...is there ever a reasonable situation where this array has length more than 1? IDK if we should log an info message or something to indicate that something unusual is going on? 🤔

If this were scip-ruby, I would have added an assert which would only be triggered in DEBUG builds, not sure if there is an equivalent in JS/TS land.

valerybugakov: Method overloads and namespace declarations spread across multiple files will cause this array to have more than one entry. E.g.,

https://github.com/sourcegraph/scip-typescript/blob/6d69827373bebe01b07611a9b142e3cc0f892ec7/snapshots/output/pure-js/src/main.js#L24-L27

src/FileIndexer.ts
valerybugakov: It happens because we pass this method as a callback to the `.map` function. It will be called without `this` context.

```tsx
sym.declarations
  .map(declaration => this.scipSymbol(declaration).value)
```

For example:

```tsx
// `this` is used as a context for the method called.
this.scipSymbol(node) 

// the `scipSymbol` call doesn't have any predefined execution context. The global context will be used.
myArray.map(this.scipSymbol) 
```

Rules behind this:

> Determining the `this` binding for an executing function requires finding the direct call-site of that function. Once examined, four rules can be applied to the call-site, in *this* order of precedence:
>  1. Called with `new`? Use the newly constructed object.
>  2. Called with `call` or `apply` (or `bind`)? Use the specified object.
>  3. Called with a context object owning the call? Use that context object.
>  4. Default: `undefined` in `strict mode`, global object otherwise.

package.json
```diff
@@ -11,8 +11,8 @@
     "tslint": "tslint -p tsconfig.json --format stylish",
     "eslint": "eslint --cache '**/*.ts?(x)'",
     "build": "node ./node_modules/typescript/bin/tsc -b .",
-    "test": "uvu -r ts-node/register --ignore dist",
-    "update-snapshots": "uvu -r ts-node/register --ignore dist --update-snapshots",
+    "test": "TS_NODE_TRANSPILE_ONLY=true uvu -r ts-node/register --ignore dist",
```
olafurpg, line 14: I added new `test-fast` and `update-snapshots-fast` scripts. I prefer to get the type errors in the terminal because typechecking is reasonably fast in this codebase and debugging runtime exceptions is slower than debugging type errors.

src/FileIndexer.ts
olafurpg: In this codebase, we generally try to use for loops over map/flatMap so this change should not be needed.

src/FileIndexer.ts
```diff
@@ -378,7 +395,11 @@ export class FileIndexer {
       return this.cached(node, this.scipSymbol(node.parent))
     }
 
-    if (ts.isImportSpecifier(node) || ts.isImportClause(node)) {
+    if (
+      ts.isImportSpecifier(node) ||
+      ts.isImportClause(node) ||
+      ts.isNamespaceImport(node)
```
olafurpg, line 401: Is this change related to de-duplication? It would be nice to isolate this change and have a diff in the snapshot tests to better understand the exact motivation for this change.

src/FileIndexer.ts
```diff
@@ -608,6 +629,25 @@ export class FileIndexer {
...
+   * @param sym ts.Symbol with declarations to process.
+   * @returns an array of unique non-empty SCIP symbol values extracted
+   * from ts.Symbol declarations.
+   */
+  private uniqueDeclarationSymbolValues(sym?: ts.Symbol): string[] {
+    if (!sym?.declarations) {
+      return []
+    }
+
+    return Array.from(
```
olafurpg, line 643: This expression is fairly expensive considering it's in a very hot loop and that we can achieve a similar results with a single if condition, as is done in
https://github.com/sourcegraph/scip-typescript/pull/194

It's very important that we try to keep performance parity with `tsc -b` and the fewer redundant collections we allocate the better.

src/FileIndexer.ts
```diff
@@ -97,42 +111,48 @@ export class FileIndexer {
     return symbol
   }
 
-  private visitSymbolOccurrence(node: ts.Node, sym: ts.Symbol): void {
-    const range = Range.fromNode(node).toLsif()
-    let role = 0
-    const isDefinition = this.declarationName(node.parent) === node
-    if (isDefinition) {
-      role |= scip.scip.SymbolRole.Definition
+  private visitDefinition(node: ts.Node, sym: ts.Symbol): void {
```
olafurpg, line 114: I'm not convinced this refactoring is an improvement in maintainability. In the future, I want to add more symbol roles such as import and write and then the `symbol_roles` bitmask will need to be updated in two different places.

src/FileIndexer.ts
```diff
@@ -65,7 +65,21 @@ export class FileIndexer {
     if (ts.isIdentifier(node) || ts.isStringLiteralLike(node)) {
       const sym = this.getTSSymbolAtLocation(node)
       if (sym) {
-        this.visitSymbolOccurrence(node, sym)
+        /**
```
olafurpg, line 68: I agree it's not super clear why `this.declarationName(node.parent) === node` identifies a definition. To improve readability, I'd prefer to add a function `private isDefinition(...): boolean` with a docstring than adding an inline comment like this.

olafurpg (commented):
@valerybugakov Thank you for this contribution! There are several independent changes in this PR that make it harder to review than necessary. I opened a separate PR https://github.com/sourcegraph/scip-typescript/pull/194 to demonstrate how the issue can be fixed with a smaller diff. In the future, I encourage you to try and open separate PRs for independent changes. It's especially important that functional changes like the new `ts.isNamespaceImport(node)` condition are documented with a minimized diff in the git history.

package.json
valerybugakov: Sounds good! To clarify the reasoning behind this: while iterating on changes locally, I often have type errors because I want to test some logic but don't want to invest time into making Typescript happy about it because there's no intent to commit it. Type-checking slows down the process of trying different ideas on existing tests.

src/FileIndexer.ts
valerybugakov: https://github.com/sourcegraph/scip-typescript/pull/195

src/FileIndexer.ts
valerybugakov: https://github.com/sourcegraph/scip-typescript/pull/196

package.json
olafurpg: I'm +1 to adding these scripts to package.json as long as we keep the typechecking variants as well. For sg/sg, I appreciate that `sg start` doesn't typecheck because it's a big codebase slowing down the edit/reload/debug feedback loop. In scip-typescript, typechecking has been fast enough for my needs.

PR #195, opened 2022-10-28
Generate correct `scipSymbol` for namespace imports

valerybugakov:
Without changes introduced in this PR `local 0` SCIP symbol is generated for the `ts` import.

### Test plan

See the updated test snapshot. Comment out the `FileIndexer` changes to reproduce the previous behavior.


snapshots/input/syntax/src/import.ts
```diff
@@ -1,3 +1,4 @@
+import * as ts from 'typescript'
```
varungandhi-src: nit: Could you define a namespace in a different file (similar to `./class` and `./enum` below) and use that here? We try to make tests as self-contained as possible, whereas using `SyntaxKind` and `ArrayType` kinda' goes against that. If one is looking at the test by itself, it isn't obvious whether the `'typescript'`/`ts` is load-bearing or if it is more general.

varungandhi-src:
> See the updated test snapshot. Comment out the FileIndexer changes to reproduce the previous behavior.

One thing I like to sometimes do here is first commit the test (and run update-snapshots -- so that tests are still passing) and then commit the fix with the updated snapshot. The commits are squashed when merging anyways, but the PR becomes slightly nicer to read (this commit is the fix, and here is how the snapshot output changed because of it). The flip side is that is becomes a bit more finicky if you want to go back and edit the test later. I've sometimes done this for PRs in scip-ruby this way; it also makes it easier to go back on look at older PRs even if it was only self-reviewed. For example:
- https://github.com/sourcegraph/scip-ruby/pull/107/commits
- https://github.com/sourcegraph/scip-ruby/pull/83/commits

I haven't seen other people do this though, so I'm not going to ask that you do the same, but just mentioning it as one option. 😅

valerybugakov:
@varungandhi-src, I like your approach! Here's [the line](https://github.com/sourcegraph/scip-typescript/pull/195/commits/af4bb95001872116df3d7e20b8d165002a90d2b6#diff-bfcd5572fd5d4d9227e5d350db9a53f9db2d91a64998df7a486b1b8b0038d96dR37) with the incorrectly generated symbol. And here's [the update](https://github.com/sourcegraph/scip-typescript/pull/195/commits/e62facf6dfc36ac141c40162a206fb416ca2483e#diff-bfcd5572fd5d4d9227e5d350db9a53f9db2d91a64998df7a486b1b8b0038d96dR37) that fixes it.

varungandhi-src (approved):

snapshots/output/syntax/src/import.ts
```diff
@@ -33,10 +33,10 @@
 //              ^ reference syntax 1.0.0 src/`enum.ts`/Enum#A.
       newFunction() +
 //    ^^^^^^^^^^^ reference syntax 1.0.0 src/`function.ts`/newFunction().
-      ts.SyntaxKind.ArrayType
-//    ^^ reference typescript 4.8.4 lib/`typescript.d.ts`/ts/
-//       ^^^^^^^^^^ reference typescript 4.8.4 lib/`typescript.d.ts`/ts/SyntaxKind#
-//                  ^^^^^^^^^ reference typescript 4.8.4 lib/`typescript.d.ts`/ts/SyntaxKind#ArrayType.
+      namespace.a.value
+//    ^^^^^^^^^ reference local 0
```
olafurpg: Nice find 👍🏻

olafurpg (approved):
This looks great. Thank you much @valerybugakov!

olafurpg:
> One thing I like to sometimes do here is first commit the test (and run update-snapshots so that tests are still passing) and then commit the fix with the updated snapshot. 

@varungandhi-src I regularly do this myself. It's a very helpful technique to verify that your diff actually fixes the issue.

PR #196, opened 2022-10-28
Extract `isDefinition` helper

valerybugakov:
This PR improves the readability of the `isDefinition` logic. Follow-up on [this comment](https://github.com/sourcegraph/scip-typescript/pull/192/files#r1006949009). 

### Test plan

Tests are green, and the comment on the `isDefinition` function looks ok.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,5 +10,7 @@ module.exports = {
   ignorePatterns: ['temp', 'scip.ts', 'snapshots'],
   rules: {
     'no-sync': 'off',
+    'jsdoc/check-indentation': 'off',
+    'class-methods-use-this': 'error',
   },
 }
--- a/src/FileIndexer.ts
+++ b/src/FileIndexer.ts
@@ -100,8 +100,8 @@ export class FileIndexer {
   private visitSymbolOccurrence(node: ts.Node, sym: ts.Symbol): void {
     const range = Range.fromNode(node).toLsif()
     let role = 0
-    const isDefinition = this.declarationName(node.parent) === node
-    if (isDefinition) {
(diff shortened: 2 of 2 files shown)


.eslintrc.js
```diff
@@ -10,5 +10,7 @@ module.exports = {
   ignorePatterns: ['temp', 'scip.ts', 'snapshots'],
   rules: {
     'no-sync': 'off',
+    'jsdoc/check-indentation': 'off',
+    'class-methods-use-this': 'error',
```
valerybugakov, line 14: Follow-up on [this comment](https://github.com/sourcegraph/scip-typescript/pull/192#discussion_r1006783990) by @varungandhi-src. [This rule](https://eslint.org/docs/latest/rules/class-methods-use-this) forbids class methods without instance context usage. @olafurpg, let me know what you think about that!

src/FileIndexer.ts
```diff
@@ -707,3 +681,46 @@ function isEqualArray<T>(a: T[], b: T[]): boolean {
   }
   return true
 }
+
+function declarationName(node: ts.Node): ts.Node | undefined {
```
valerybugakov, line 685: Extracted into an independent function to satisfy the `class-methods-use-this` ESLint rule.

.eslintrc.js
olafurpg: This is really nice. Big +1 from me

src/FileIndexer.ts
olafurpg: It's a big improvement in readability to have a toplevel function instead of an instance method.

olafurpg (approved):
LGTM 👍🏻 Very nice!

PR #206, opened 2022-11-11
Fix bug where an interface definition had a conflicting symbol with a const in the same file

olafurpg:
Reported via Slack https://sourcegraph.slack.com/archives/CHXHX7XAS/p1668157176918589?thread_ts=1668128629.962219&cid=CHXHX7XAS

### Test plan
See the snapshot test for a minimized reproduction. I also manually verified the bug has been fixed by indexing microsoft/vscode with a local build and uploading the index to https://sourcegraph.com/github.com/microsoft/vscode@9984da1/-/blob/src/vs/workbench/services/workspaces/common/workspaceEditing.ts?L13:18#tab=implementations_typescript

### DIFF
--- a/snapshots/output/syntax/src/conflicting-const-interface.ts
+++ b/snapshots/output/syntax/src/conflicting-const-interface.ts
@@ -4,6 +4,12 @@
 //             ^^^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`conflicting-const-interface.ts`/ConflictingConst.
 //             documentation ```ts\ninterface ConflictingConst\n```
   export interface ConflictingConst {}
-//                 ^^^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`conflicting-const-interface.ts`/ConflictingConst.
+//                 ^^^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`conflicting-const-interface.ts`/ConflictingConst#
 //                 documentation ```ts\ninterface ConflictingConst\n```
+  export class ImplementsConflictingConst implements ConflictingConst {}
+//             ^^^^^^^^^^^^^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`conflicting-const-interface.ts`/ImplementsConflictingConst#
+//             documentation ```ts\nclass ImplementsConflictingConst\n```
+//             relationship implementation scip-typescript npm syntax 1.0.0 src/`conflicting-const-interface.ts`/ConflictingConst#
+//                                                   ^^^^^^^^^^^^^^^^ reference syntax 1.0.0 src/`conflicting-const-interface.ts`/ConflictingConst.
+//                                                   ^^^^^^^^^^^^^^^^ reference syntax 1.0.0 src/`conflicting-const-interface.ts`/ConflictingConst#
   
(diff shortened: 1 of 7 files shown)


snapshots/output/syntax/src/conflicting-const-interface.ts
```diff
@@ -0,0 +1,9 @@
+  export const ConflictingConst = 42
+// definition syntax 1.0.0 src/`conflicting-const-interface.ts`/
+//documentation ```ts\nmodule "conflicting-const-interface.ts"\n```
+//             ^^^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`conflicting-const-interface.ts`/ConflictingConst.
+//             documentation ```ts\ninterface ConflictingConst\n```
+  export interface ConflictingConst {}
+//                 ^^^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`conflicting-const-interface.ts`/ConflictingConst.
```
olafurpg: The expected behavior
```suggestion
//                 ^^^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`conflicting-const-interface.ts`/ConflictingConst#
```

src/FileIndexer.ts
```diff
@@ -37,6 +37,10 @@ export class FileIndexer {
     this.workingDirectoryRegExp = new RegExp(options.cwd, 'g')
   }
   public index(): void {
+    // Uncomment below if you want to skip certain files for local development.
+    // if (!this.sourceFile.fileName.includes('conflicting-const')) {
+    //   return
+    // }
```
varungandhi-src, line 43: Why add these special comments for particular test cases? We could potentially add a test filter regex or something if that's desired, this seems a little strange.

src/FileIndexer.ts
```diff
@@ -104,7 +108,10 @@ export class FileIndexer {
     if (isDefinitionNode) {
       role |= scip.scip.SymbolRole.Definition
     }
-    for (const declaration of sym?.declarations || []) {
+    const declarations = isDefinitionNode
+      ? [node.parent] // Don't emit ambiguous definition at definition-site
+      : sym?.declarations || []
```
varungandhi-src: This comment doesn't properly explain what is going on. Why does `sym?.declarations` lead to ambiguity? Can we trust it for non-definition nodes? It seems like there is something unusual going on with the AST structure here which needs more explanation.

src/FileIndexer.ts
olafurpg: I haven't figured out yet how to implement this kind of filtering from the test framework so I prefer to keep it as a comment until we figure out a better solution.

src/FileIndexer.ts
olafurpg: You can reproduce the ambiguous definition with tsserver

```
export const Conflict = 42
interface Conflict {}
          ^^^^^^^^ Go to definition here shows two results: the const and the interface
```

I updated the comment to give this example and link to this PR. 

> Can we trust it for non-definition nodes?

`sym.declarations` is what tsserver uses so I think we need to trust it except for odd-cases like this one where it seems like the API has a bug.

Strum355 (approved):

PR #209, opened 2022-11-28
Add support for private identifiers

### LINKED ISSUE #208: Doesn't pick `#` fields
scip-typescript doesn't seem to pick up fields that use the private `#` modifier

For example, [`#getServer`](https://sourcegraph.com/gitlab.com/gitlab-org/gitlab-vscode-extension@2f05c986f2ce767022578e613348469090207dcd/-/blob/src/completion/gitlab_code_completion_provider.ts?L45&subtree=true#tab=def) only has search-based refs.

### LINKED ISSUE #280: Smarter auto-inference for JavaScript only repos
In some cases, we can still run into the problem where the current auto-inferred tsconfig.json doesn't actually index any JS files. Here's an auto-indexing log showing that:

```
+ '[' -n 2048 ']'
+ export NODE_OPTIONS=--max-old-space-size=2048
+ NODE_OPTIONS=--max-old-space-size=2048
+ scip-typescript index --infer-tsconfig
error: no files got indexed. To fix this problem, make sure that the TypeScript projects ["/data"] contain input files or reference other projects.
```

Can we be smarter in inferring the tsconfig.json to avoid this in more cases?

olafurpg:
Fixes #208

### Test plan
See new snapshot test.

### DIFF
--- a/snapshots/input/syntax/src/ClassWithPrivate.ts
+++ b/snapshots/input/syntax/src/ClassWithPrivate.ts
@@ -1,0 +1,22 @@
+export class ClassWithPrivate {
+  #privateField
+  #privateFieldWithInitializer = 42
+
+  #privateMethod() {
+    this.#privateField = 'private field'
+    return this.#privateField
+  }
+
+  static #privateStaticField
+  static #privateStaticFieldWithInitializer = 42
+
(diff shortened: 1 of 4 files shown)


tjdevries (approved):
LGTM

PR #248, opened 2023-04-20
Infer symbol relationship for object literals property in return statement

olafurpg:
Previously, scip-typescript didn't infer correct relationships between the object literel property and interface property in return statements like below:

```ts
interface Configuration {
  property: number
}

export function returnStatement(): Configuration {
  return {
    property: 42,
  }
}
```

This commit fixes this specific issue. This commit also adds a test case for a situation where scip-typescript doesn't infer the relationship today but tsserver does. We should fix that in the future.

### Test plan

See updated snapshot tests.

### DIFF
--- a/snapshots/input/syntax/src/infer-relationship.ts
+++ b/snapshots/input/syntax/src/infer-relationship.ts
@@ -2,7 +2,49 @@ interface Configuration {
   property: number
 }
 
+function random(): number {
+  return Math.random()
+}
 export function returnStatement(): Configuration {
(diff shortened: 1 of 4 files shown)


keynmol (approved):

olafurpg:
Realized the previous logic failed as soon as you didn't have a toplevel return statement. I have a half-baked idea how to clean up the logic so that we pass down the inferred type while traversing the AST avoiding the need to look up parents (which is typically a sign of a broken abstraction IME).

PR #250, opened 2023-04-24
Add definitions and references for constructors

### LINKED ISSUE #249: Emit occurrences for class constructors and their usage sites
Currently a constructor of a class is marked as a reference of that class and any usage sites (`new MyClass()`) are marked as occurrences for `MyClass`, but neither of these are marked as occurrences of the `<constructor>` method as they should be.

SuperAuguste:
Closes #249

This replicates VSCode's gotodef.

### Test plan

Tested locally with a snapshot, still need to test edge cases + implement and test inherited constructors.


src/FileIndexer.ts
```diff
@@ -38,9 +38,9 @@ export class FileIndexer {
   }
   public index(): void {
     // Uncomment below if you want to skip certain files for local development.
-    // if (!this.sourceFile.fileName.includes('infer-relationship')) {
-    //   return
-    // }
+    if (!this.sourceFile.fileName.includes('constructor')) {
```
olafurpg: Let's uncomment this and re-run the snapshots

src/main.test.ts
```diff
@@ -35,9 +35,9 @@ interface PackageJson {
 }
 for (const snapshotDirectory of snapshotDirectories) {
   // Uncomment below if you want to skip certain tests for local development.
-  // if (!snapshotDirectory.includes('syntax')) {
-  //   continue
-  // }
+  if (!snapshotDirectory.includes('syntax')) {
```
olafurpg: Also uncomment here

src/FileIndexer.ts
```diff
@@ -480,7 +516,9 @@ export class FileIndexer {
         return undefined
       }
       const signatureDeclaration: ts.SignatureDeclaration | undefined =
-        ts.isFunctionDeclaration(declaration)
+        ts.isConstructorDeclaration(node)
```
olafurpg: Minor: it's probably about time to move this ternary into a separate function

snapshots/input/syntax/src/constructor.ts
```diff
@@ -0,0 +1,7 @@
+export class Constructor {
+  constructor(public readonly property: number) {}
+}
+
+export function useConstructor(): Constructor {
```
olafurpg: Can we add a test case for this syntax `new a.A()`? We have examples of this in the codebase 

```
        new scip.scip.Occurrence({
          range,
          symbol: scipSymbol.value,
          symbol_roles: role,
        })
```

I think it's OK to assume `node.expression` will only be an identifier or member selection (can't remember what tsc calls it: `a.b` expressions).

src/FileIndexer.ts
```diff
@@ -76,6 +77,31 @@ export class FileIndexer {
         this.visitSymbolOccurrence(node, sym)
       }
     }
+    
+    if (
```
olafurpg: Nit: I'd prefer to move this logic into `getTSSymbolLocation` and `visitSymbolOccurrence` by looking at the parents to determine if the identifier is part of a new expression. We already emit an occurrence over the identifier, it's just the wrong symbol.

My undocumented goal with the `visit()` function is to have it as small as possible by exclusively performing the traversal.

src/FileIndexer.ts
```diff
@@ -76,6 +77,31 @@ export class FileIndexer {
...
+      ts.isNewExpression(node) &&
+      ts.isIdentifier(node.expression)
+    ) {
+      const sym = this.getTSSymbolAtLocation(node.expression)
+      const declaration = sym?.declarations?.at(0);
+      if (declaration && ts.isClassDeclaration(declaration)) {
+        for (const member of declaration.members) {
```
olafurpg: I think it's better to just do `ScipSymbol.global(classDeclarationSymbol, methodDescriptor("<constructor>"))` 🤔 It's probably not a performance issue for cases when the constructor is the first member of this list but it could become a performance issue when the constructor doesn't exist (for whatever reason) and we exhaustively iterate through a very large number of members (TypeScript types can become very big with union types!).

src/FileIndexer.ts
```diff
@@ -305,6 +334,13 @@ export class FileIndexer {
           }
         }
       })
+    } else if (ts.isConstructorDeclaration(declaration)) {
+      relationships.push(
```
olafurpg: Can you elaborate on the motivation for this relationship? My understanding is that it's redundant since we're already emitting a definition occurrence.

CleanShot 2023-04-25 at 10 55 06@2x

Definition relationships are only useful for symbols that are missing a definition occurrence.

olafurpg (changes requested):
Impressive first PR @SuperAuguste! This is definitely a tricky issue to fix and I see you've already become familiar with the tsc APIs 😄 

It's mostly just me who has worked in this codebase so far and there are many undocumented conventions and assumptions. Don't hesitate to push back if some of these comments are too nitpicky, I prefer to have more relaxed conventions if it makes it easier for more people to contribute (rather than strictly enforcing some undocumented style making it a drudge to contribute).

src/FileIndexer.ts
SuperAuguste: Oops thanks for letting me know 😅 I'll rewrite this

src/FileIndexer.ts
SuperAuguste: This makes sense, but what if there's no constructor? Then we'd just be pointing to nothing which isn't the behavior we want :(

SuperAuguste:
CI failing due to different TypeScript versions :/

I'm kind of lost here, not sure why they're different to be frank.

snapshots/package.json
```diff
@@ -8,8 +8,8 @@
     "input/*"
   ],
   "dependencies": {
-    "react": "18.0.0",
-    "@sourcegraph/tsconfig": "4.0.1"
+    "@sourcegraph/tsconfig": "4.0.1",
```
olafurpg: Was this change intentional?

src/FileIndexer.ts
olafurpg: You're right! We need to fallback to the class if the constructor doesn't exist. We can add a minimized test case to reproduce it

```ts
export class NoConstructor {
  property: number
}

export function useNoConstructor() {
  return new NoConstructor()
}
```

We will need the original for loop to detect if the class has a constructor but I propose we cache the result in a `hasConstructor: Map<ScipSymbol, boolean>` that lives in `ProjectIndexer` (next to `symbolCache`). Without the cache we risk iterating through all the members on every new expression, and I can imagine pathological cases (for example, generated code) where this iteration could become expensive.

olafurpg (changes requested):
This PR is really shaping up! Only blocking comments are

- Add back your constructor detection logic, but with a cache
- Undo changes in yarn.lock, I suspect this might fix the failing CI

snapshots/package.json
SuperAuguste: Uh running yarn install just changed the order of the packages, I don't think this affects behavior but yeah this wasn't intentional :(

SuperAuguste:
> Undo changes in yarn.lock, I suspect this might fix the failing CI

The current package.json seems to be invalid due to a peer resolution issue (I ran `yarn install --immutable --immutable-cache` to not modify the `yarn.lock`) so my install won't go through. How do you install the deps for this repo without modifying the package.json/lockfile? :)

SuperAuguste:
Figured out the issue - the yarn version used here is a few years behind it seems, so I installed v1.22.19 and everything worked out!

Development.md
```diff
@@ -1,5 +1,7 @@
 # Developing scip-typescript
 
+Please note that the yarn version used by CI is `v1.22.19` - you should use this version as well to prevent lockfile conflicts.
```
olafurpg, line 3: Followup https://github.com/sourcegraph/scip-typescript/issues/251

snapshots/output/syntax/src/constructor.ts
```diff
@@ -0,0 +1,83 @@
...
+//             documentation ```ts\nclass NoConstructor\n```
+    property: number
+//  ^^^^^^^^ definition syntax 1.0.0 src/`constructor.ts`/NoConstructor#property.
+//  documentation ```ts\n(property) property: number\n```
+  }
+  
+  export function useNoConstructor() {
+//                ^^^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`constructor.ts`/useNoConstructor().
+//                documentation ```ts\nfunction useNoConstructor(): NoConstructor\n```
+    return new NoConstructor()
```
olafurpg, line 80: Nice! 🤩

src/FileIndexer.ts
```diff
@@ -474,17 +512,24 @@ export class FileIndexer {
...
-    const signature = (): string | undefined => {
+    const asSignatureDeclaration = (
+      node: ts.Node,
+      sym: ts.Symbol
+    ): ts.SignatureDeclaration | undefined => {
       const declaration = sym.declarations?.[0]
       if (!declaration) {
         return undefined
       }
-      const signatureDeclaration: ts.SignatureDeclaration | undefined =
```
olafurpg, line 482: I hadn't realized this was already inside a function 🤦🏻 The new code looks good though 👍🏻

olafurpg (approved):
Glorious 😍 Thank you @SuperAuguste!

olafurpg:
> . How do you install the deps for this repo without modifying the package.json/lockfile? :)


I normally run `yarn` without a subcommand

PR #253, opened 2023-04-28
Fix object literal references

### LINKED ISSUE #252: Emit references instead of definitions for properties on object literals that implement an interface
Currently, scip-typescript emits a definition for `property: 41` in cases like below
```
  interface Configuration {
    property: number
//  ^^^^^^^^ definition syntax 1.0.0 src/`infer-relationship.ts`/Configuration#property.
//  documentation ```ts\n(property) property: number\n```
  }
        return {
          property: 41,
//        ^^^^^^^^ definition syntax 1.0.0 src/`infer-relationship.ts`/property3:
//        documentation ```ts\n(property) property: number\n```
//        relationship implementation reference scip-typescript npm syntax 1.0.0 src/`infer-relationship.ts`/Configuration#property.
        }
```

Minimized reproduction in the tests
https://github.com/sourcegraph/scip-typescript/blob/main/snapshots/output/syntax/src/infer-relationship.ts#L29

We should change the behavior to emit a reference occurrence to `src/`infer-relationship.ts`/Configuration#property.` instead.

SuperAuguste:
Closes #252

### Test plan

Snapshots


olafurpg:
While reviewing this PR, I realized that the codebase doesn't have good enough infrastructure to fix this issue cleanly. The problem is that `inferredTypeOfObjectLiteral` is sort of a hack, it's difficult to reason about and it's difficult to extend to support some more advanced use-cases.


I took a stab at implementing another approach that tracks which expressions have explicitly ascribed types in the `visit` function. The benefit of this approach is that we avoid the recursive traversal up the parent chain to figure out if the object literal has an explicitly assigned type. https://github.com/sourcegraph/scip-typescript/pull/253 

That PR doesn't yet handle all cases where we should emit references for properties. Specifically, the `returnStatementInsideArgumentExpression` snapshot test is not yet handled and it's a more tricky to deal with than the other cases since we need to unwind the function argument's type. Would you be interested in taking a stab at handling that case?

PR #256, opened 2023-05-01
Emit references instead of definitions for object literal properties

### LINKED ISSUE #252: Emit references instead of definitions for properties on object literals that implement an interface

olafurpg:
Fixes #252.

Previously, scip-typescript emitted definitions for properties of object literal expressions that got inferred to a concrete type. This was problematic since doing "Find references" didn't behave like it does in VS Code, where it shows the object literal property together with the type's member.

For example, consider
```ts
interface Config { property: number }
const x: Config = { property: 42 }
```

After this PR, the `property` in `property: 42` becomes a regular reference to `Config.property` instead of being its own definition. This means that "Find references" now works like it does in VS Code.

### Test plan

See updated snapshots.


snapshots/output/syntax/src/infer-relationship.ts
```diff
@@ -76,60 +88,120 @@
...
       const incremented = number + 1
-//          ^^^^^^^^^^^ definition local 15
+//          ^^^^^^^^^^^ definition local 24
 //          documentation ```ts\nvar incremented: number\n```
-//                        ^^^^^^ reference local 12
+//                        ^^^^^^ reference local 21
       return {
         property: incremented,
-//      ^^^^^^^^ definition syntax 1.0.0 src/`infer-relationship.ts`/property7:
+//      ^^^^^^^^ definition syntax 1.0.0 src/`infer-relationship.ts`/property0:
```
olafurpg: @SuperAuguste I don't know yet how to handle the case here but I think we can do it somehow by adding more entries to the `assignedType` map. Would you be interested in taking a look?

snapshots/output/syntax/src/inheritance.ts
```diff
@@ -12,7 +12,7 @@
 //  documentation ```ts\n(property) property: string\n```
     interfaceMethod(): string
 //  ^^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`inheritance.ts`/Superinterface#interfaceMethod().
-//  documentation ```ts\n(method) interfaceMethod() => string\n```
+//  documentation ```ts\n(property) interfaceMethod: () => string\n```
```
olafurpg: I have no clue why this changed. I don't think it's a big problem but it is concerning that I can't explain why it changed 🤔

snapshots/output/syntax/src/inheritance.ts
```diff
@@ -136,11 +132,11 @@
     consumesInterface({
 //  ^^^^^^^^^^^^^^^^^ reference syntax 1.0.0 src/`inheritance.ts`/consumesInterface().
       interfaceMethod: (): string => 'inferred',
-//    ^^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`inheritance.ts`/interfaceMethod1:
+//    ^^^^^^^^^^^^^^^ definition syntax 1.0.0 src/`inheritance.ts`/interfaceMethod0:
```
olafurpg: This case is still unhandled. It's a simpler than `returnStatementInsideArgumentExpression` so it could be a good one to look at first

snapshots/output/syntax/src/structural-type.ts
```diff
@@ -65,7 +65,7 @@
 //     ^^^^^^^^^^^^ definition syntax 1.0.0 src/`structural-type.ts`/FeatureFlags#
 //     documentation ```ts\ntype FeatureFlags\n```
 //                      ^^^^^^^^ definition syntax 1.0.0 src/`structural-type.ts`/FeatureFlags#typeLiteral13:darkMode.
-//                      documentation ```ts\n(property) darkMode: () => void\n```
+//                      documentation ```ts\n(property) darkMode: true\n```
```
olafurpg: This is a serious regression! Need to look into it better, I think I know what's going on, and fixing this will probably fix the other `method/property` regression

snapshots/output/syntax/src/inheritance.ts
olafurpg: Fixed now.

snapshots/output/syntax/src/infer-relationship.ts
SuperAuguste: Yep, definitely interested!

snapshots/output/syntax/src/inheritance.ts
SuperAuguste: I'll look at this first then :)

SuperAuguste:
Resolved everything here; this is some of the hackiest stuff I've put together in a while so heads up - it seems to work though :)

olafurpg:
Finally got a moment to look again into this PR and fixed one big remaining blocker to merge this PR. This PR is big and I want to make the changes low risk to merge as possible. I will need a few more iterations before it's safe to merge but I'm eager to get it through the finish line since it fixes a ton of very common navigation quirks.

olafurpg:
@valerybugakov @SuperAuguste this PR is ready for review now. I cleaned up the commit history so that you can only review the differences in the indexed output by looking at the diff of the last commit. The logic in `updateAssignedTypeMap` is tricky to reason about because it's inherently complicated to mirror TypeScript typing rules. I did my best to keep the code clean, and I think the code is much cleaner now compared to how we inferred implementation relationships before this PR.

rodion-m:
What prevents us from merging this?

olafurpg:
@rodion-m main reason this hasn't been merged yet is because I'm personally working on other projects and haven't found the time to do the rebase, find reviewers, cut a release and communicate the changes (along with addressing followup feedback). I think the changes in this PR are a significant improvement to the TypeScript/JavaScript navigation experience. I'd love to see this land eventually but don't have an ETA for it at the moment.

kritzcreek:
Superseeded by https://github.com/sourcegraph/scip-typescript/pull/373

Thanks a lot for all the great test cases in this PR!

PR #271, opened 2023-07-11
Skip indexing large files above 1mb, add option `--max-file-byte-size`

olafurpg:
Previously, scip-typescript indexed all files regardless of file size. This could result in scip-typescript stalling progress to index very large files that were (frequently) auto-generated. This commit changes the default behavior to skip indexing files that are larger than 1mb, and makes this threshold configurable via the new `--max-file-byte-size` flag.

### Test plan

- New unit tests for the parse function
- Manually tested in the sourcegraph/sourcegraph repo with the command `❯ ts-node ~/dev/sourcegraph/scip-typescript/src/main.ts index --pnpm-workspaces --max-file-byte-size 500kb`. Printed out skipped files and confirmed they all exceeded 500kb.

### DIFF
--- a/src/FileIndexer.ts
+++ b/src/FileIndexer.ts
@@ -15,6 +15,7 @@ import {
 } from './Descriptor'
 import { Input } from './Input'
 import { Packages } from './Packages'
+import { formatByteSizeAsHumanReadable } from './parseHumanByteSizeIntoNumber'
 import { Range } from './Range'
 import * as scip from './scip'
 import { ScipSymbol } from './ScipSymbol'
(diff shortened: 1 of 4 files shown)


src/FileIndexer.ts
```diff
@@ -42,6 +42,16 @@ export class FileIndexer {
     // if (!this.sourceFile.fileName.includes('constructor')) {
     //   return
     // }
+
+    if (
+      this.options.maxFileByteSizeNumber &&
+      Buffer.from(this.sourceFile.getText()).length >
+        this.options.maxFileByteSizeNumber
+    ) {
+      // Skip files that exceed the --max-file-byte-size threshold.
```
keynmol: Do we want to log that the file was skipped?

olafurpg: Good idea. Updated the implementation to log which files are skipped.

olafurpg: Looks like this

```
info: skipping file '/Users/olafurpg/dev/sourcegraph/sourcegraph/client/web/src/enterprise/sentinel/LandingImage.ts' because it has byte size 110.621kb that exceeds the maximum threshold 100kb. If you intended to index this file, use the flag --max-file-byte-size to configure the maximum file size threshold.
```

keynmol (approved):

PR #293, opened 2023-12-15
Implement enclosing ranges

Titou325:
Implements basic `enclosing_ranges` for JS/TS.
The following are implemented:

* Named arrow function declaration,
* Function declaration,
* Class declaration,
* Method declaration,
* TS Interface declaration

To facilitate snapshot handling, I rebased the current snapshot formatter on the one used for Python. These two should be put in common at a later stage.

The implementation itself is very straightforward, and I believe can easily be extended to support other features.

Related to #278 

### Test plan

Added the following snapshots, with their input and output. The output was visually confirmed.

* Added `enclosing-range` snapshot to test JS support,
* Added `enclosing-range-ts` snapshot to test TS support

PR #306, opened 2023-12-25
test: Re-use snapshot formatting from scip-python

varungandhi-src:
The feature implementation PR #293 changes the snapshot formatting for a large
number of snapshots, making it harder to verify that there are no unexpected
changes due to the changes in the code. This patch separates out the
snapshot formatting so that it is clearer what the mapping from
code changes -> snapshot output changes is.

### Test plan

Covered by existing snapshot tests


varungandhi-src:
cc @Titou325

valerybugakov (approved):

PR #343, opened 2024-05-15
Fix SCIP uploads in CI

varungandhi-src:
Fixes GRAPH-584

### Test plan

Check that uploads work on this PR


keynmol (approved):
LGTM, need to run prettier

PR #350, opened 2024-06-04
Fix path normalize check for Windows

bulldy80:
`path.normalize` on Windows returns paths with back-slashes (\\) instead of forward -slashes (/), which causes the check here to fail.

This change fixes the issue by replacing all the back-slashes with forward-slashes.

### Test plan

Ran indexer on Windows and verified the contents of the index file are as expected.


olafurpg (approved):
LGTM 👍🏻 Thank you!

Issue #360, opened 2024-07-24
`SymbolInformation` does not contain `SymbolInformation.Kind kind` field. 


soohoonc:
When running `scip-typescript index` and viewing the outputs, none of the symbols have the `kind` field outlined in the `scip.proto` schema. Thought this information would be included in the index output, wanted to use it for some of the stuff I was building.
```[tasklist]
### Tasks
```

varungandhi-src:
Not all SCIP indexers emit all fields. We do not currently have an immediate use case for this field at Sourcegraph, so we are unlikely to implement it. However, if you'd like to submit a PR to add this feature, we can review it.

PR #361, opened 2024-07-24
Add `kind` and `display_name` to `SymbolInformation`

soohoonc:
Started adding support for the `kind` and `display_name` field for `SymbolInformation`.

### Test plan
Currently tested on couple of personal typescript repos, may need to update snapshot tests (not totally sure what the process here is like)

### DIFF
--- a/src/FileIndexer.ts
+++ b/src/FileIndexer.ts
@@ -82,6 +82,8 @@ export class FileIndexer {
       new scip.scip.SymbolInformation({
         symbol: symbol.value,
         documentation: ['```ts\nmodule "' + moduleName + '"\n```'],
+        kind: scip.scip.SymbolInformation.Kind.Module,
+        display_name: moduleName,
       })
     )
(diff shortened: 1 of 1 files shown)


soohoonc:
mapping from typescript types to symbolInformation is incomplete/may be inaccurate, will work on this a bit more.

varungandhi-src:
Sorry, just saw this PR after https://github.com/sourcegraph/scip-typescript/issues/360

We should add a magic comment to display the kinds and symbol names in a specific file, and add new snapshotting logic for that. After that, you can update the snapshots. https://github.com/sourcegraph/scip-typescript/blob/main/Development.md

Existing snapshots should be unaffected, only the new file containing this magic comment should be affected.

Example of how we do this in scip-clang. https://sourcegraph.com/github.com/sourcegraph/scip-clang/-/blob/test/index/docs/docs.cc?L1-2

PR #373, opened 2024-09-23
fix: resolve property assignments in object literals

kritzcreek:
Simpler alternative to #256

This follows the implementation I found in the TypeScript goToDefinition service https://github.com/microsoft/TypeScript/blob/88809467e8761e71483e2f4948ef411d8e447188/src/services/goToDefinition.ts#L307-L316

Leans on the TypeScript compilers `getContextualType` function to resolve object literals to the interfaces, or type definitions they are being checked against. This means we generate useful references in a lot more places, where we'd generate "dead" definitions before.

## Test plan

Ports all tests from #256 and adds some basic new ones. Check the updated snapshot output to gauge the improvement or if there are any regressions.


olafurpg (commented):
I reviewed the LSP implementation quite thoroughly and did not find `getContextualType`. It makes a ton of sense that this API exists, and I'm 100% in favor of this much simpler approach. I would try porting some of the tests from the other PR to see how it behaves.

This change is a significantly quality improvement for TypeScript navigation so I think it's absolutely worth it to get this through the finish line.

kritzcreek:
@olafurpg Could you PTAL? I inlined some more internal definitions from the TypeScript compiler to be more in-line with how they use the `getContextualType` API.

I ported all the tests from #256, and after manually inspecting the snapshot outputs for a while I think things are looking good.

snapshots/output/react/src/LoaderInput.tsx
```diff
@@ -49,7 +49,7 @@ export const LoaderInput2: React.FunctionComponent<Props> = props => {
   return <LoaderInput loading={true} key="key" children={props.children} />
 //        ^^^^^^^^^^^ reference react-example 1.0.0 src/`LoaderInput.tsx`/LoaderInput.
 //                    ^^^^^^^ reference react-example 1.0.0 src/`LoaderInput.tsx`/Props#loading.
-//                                   ^^^ reference local 10
+//                                   ^^^ reference @types/react 18.2.39 `index.d.ts`/React/Attributes#key.
```
olafurpg, line 52: 😍

snapshots/output/syntax/src/decorators.ts
```diff
@@ -0,0 +1,36 @@
...
+//  ^^^^^^^ reference @types/node 20.10.5 `console.d.ts`/`"node:console"`/global/console.
+//          ^^^ reference typescript 5.3.3 lib/`lib.dom.d.ts`/Console#log().
+//          ^^^ reference @types/node 20.10.5 `console.d.ts`/`"node:console"`/global/Console#log().
+//                                                   ^^^^^ reference syntax 1.0.0 src/`decorators.ts`/MyDecorator().(value)
+  }
+}
+
+@MyDecorator({ property: 42, property2: '42' })
+//^^^^^^^^^^^ reference syntax 1.0.0 src/`decorators.ts`/MyDecorator().
+//             ^^^^^^^^ reference syntax 1.0.0 src/`reusable-types.ts`/Numbers#property.
```
olafurpg, line 30: 👏🏻

snapshots/output/syntax/src/object-literals-arrow-function.ts
```diff
@@ -0,0 +1,88 @@
...
+  fn: (foobar: Foobar) => Foobar
+//^^ definition syntax 1.0.0 src/`object-literals-arrow-function.ts`/hasArrowFunctionParameter().(fn)
+//     ^^^^^^ definition local 1
+//             ^^^^^^ reference syntax 1.0.0 src/`object-literals-arrow-function.ts`/Foobar#
+//                        ^^^^^^ reference syntax 1.0.0 src/`object-literals-arrow-function.ts`/Foobar#
+): Foobar {
+// ^^^^^^ reference syntax 1.0.0 src/`object-literals-arrow-function.ts`/Foobar#
+  return fn({ foobar: 42 + something })
+//       ^^ reference syntax 1.0.0 src/`object-literals-arrow-function.ts`/hasArrowFunctionParameter().(fn)
+//            ^^^^^^ reference syntax 1.0.0 src/`object-literals-arrow-function.ts`/Foobar#foobar.
```
olafurpg, line 26: Woo hoo, this is such a significant leap for navigation quality. Closest  huge gap against tsserver

src/FileIndexer.ts
```diff
@@ -147,25 +147,47 @@ export class FileIndexer {
...
   }
 
+  private getDeclarationsForPropertyAssignment(
+    node: ts.Node
+  ): ts.Declaration[] | undefined {
+    const objectElement = ts_inline.getContainingObjectLiteralElement(node)
+    if (!objectElement) {
+      return
+    }
+    const contextualType = this.checker.getContextualType(objectElement.parent)
```
olafurpg, line 157: So much heavy lifting here.

src/FileIndexer.ts
```diff
@@ -690,60 +711,6 @@ export class FileIndexer {
...
   }
-
-  // Returns the "inferred" type of the provided object literal, where
-  // "inferred" is loosely defined as the type that is expected in the position
-  // where the object literal appears.  For example, the object literal in
-  // `const x: SomeInterface = {y: 42}` has the inferred type `SomeInterface`
-  // even if `this.checker.getTypeAtLocation({y: 42})` does not return
-  // `SomeInterface`. The object literal could satisfy many types, but in this
-  // particular location must only satisfy `SomeInterface`.
-  private inferredTypeOfObjectLiteral(
```
olafurpg, line 701: ❤️❤️

src/TypeScriptInternal.ts
```diff
@@ -66,3 +66,89 @@ export function getTextOfJsxAttributeName(node: ts.JsxAttributeName): string {
     ? ts.idText(node)
     : `${ts.idText(node.namespace)}:${ts.idText(node.name)}`
 }
+
+type ObjectLiteralElementWithName = ts.ObjectLiteralElement & {
+  name: ts.PropertyName
+  parent: ts.ObjectLiteralExpression | ts.JsxAttributes
+}
+export function getContainingObjectLiteralElement(
```
olafurpg, line 76: Link to original source in comment? I probably didn't do this before but it might be a good idea

olafurpg (approved):
This is glorious. Ship it!

src/TypeScriptInternal.ts
kritzcreek: Probably a good idea. I'll look them up and add the links.

src/FileIndexer.ts
kritzcreek: Yeah this one felt pretty good :D

PR #378, opened 2024-10-02
fix(deps): remove pretty-ms dependency

kritzcreek:
Removes a dependency on `pretty-ms`. It caused problems because we now produced mixed output of CommonJS and ESM modules. All we need is a 10 line function, so I've implemented it inline.

I also noticed we weren't actually running most of our tests, so I fixed that as well.

### Test plan

Check that running `node ./dist/src/main.js index` doesn't error out on an ESM module import anymore

### DIFF
--- a/src/CommandLineOptions.test.ts
+++ b/src/CommandLineOptions.test.ts
@@ -37,3 +37,5 @@ checkIndexParser(['--infer-tsconfig'], { inferTsconfig: true })
 checkIndexParser(['--no-progress-bar'], { progressBar: false })
 checkIndexParser(['--progress-bar'], { progressBar: true })
 checkIndexParser(['--no-global-caches'], { globalCaches: false })
+
+test.run()
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
   "dependencies": {
     "commander": "^12.1.0",
     "google-protobuf": "^3.21.4",
-    "pretty-ms": "^9.1.0",
     "progress": "^2.0.3",
     "typescript": "^5.6.2"
   },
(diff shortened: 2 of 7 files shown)


varungandhi-src (approved):

varungandhi-src:
> It caused problems because we now produced mixed output of CommonJS and ESM modules. All we need is a 10 line function, so I've implemented it inline.
>
> Check that running node ./dist/src/main.js index doesn't error out on an ESM module import anymore

How do people normally test this/ensure that this kind of weirdness doesn't happen?

olafurpg (approved):

src/CommandLineOptions.test.ts
```diff
@@ -37,3 +37,5 @@ checkIndexParser(['--infer-tsconfig'], { inferTsconfig: true })
 checkIndexParser(['--no-progress-bar'], { progressBar: false })
 checkIndexParser(['--progress-bar'], { progressBar: true })
 checkIndexParser(['--no-global-caches'], { globalCaches: false })
+
+test.run()
```
olafurpg, line 41: Oof

kritzcreek:
> How do people normally test this/ensure that this kind of weirdness doesn't happen?

I think you can run some kind of `npm pack` command to produce the distribution that would end up on `npm` when publishing and try to test that running that works?
Another problem is that distributing JS CLIs is a mess, because people could try to run your stuff on different node versions, or using `tsx`, or `deno`, or ...

We ran into the second case here, because I updated us to use `tsx` for running our stuff (because it supports ESM modules out of the box), but then the downstream Cody repo uses `pnpm dlx` which doesn't...

PR #389, opened 2024-12-06
Provide a single self-contained docker image, remove alpine

antonsviridov-src:
- Remove Alpine-based docker base as this forces us to use unofficial node builds: https://github.com/sourcegraph/sourcegraph/pull/2164/files, it might carry no risk, but also it isn't obvious to me what we gain from reduced image size
- Make docker build self-contained - it will bootstrap the version of scip-typescript it's built on, instead of relying on NPM publishing
- Start publishing docker images with `:latest-snapshot` on main branch builds, for easier testing and to be able to point customers to it
- Ensure we use a manfiest SHA in docker base, so that docker automatically pulls a amd64/arm64 image

### Test plan
- existing CI


.github/workflows/release.yml
```diff
@@ -39,10 +50,3 @@ jobs:
             sourcegraph/scip-typescript:${{ env.PATCH }}
             sourcegraph/scip-typescript:${{ env.MINOR }}
             sourcegraph/scip-typescript:${{ env.MAJOR }}
-      - name: Build and push
-        id: docker_build_autoindex
-        uses: docker/build-push-action@v3
-        with:
-          file: Dockerfile.autoindex
-          push: true
-          tags: sourcegraph/scip-typescript:autoindex
```
kritzcreek, line 48: Does this need a corresponding change in sg/sg to change the image we use for autoindexing?

antonsviridov-src: Yeah, I will update the SHA in sg/sg to the sha of the new container (single one).
We only refer to indexing containers by sha, not name (in default settings anyways)

kritzcreek (approved):
LGTM

PR #390, opened 2024-12-11
Remove stage id from release workflow

antonsviridov-src:
Apparently gh doesn't lint all workflows, only those that run https://github.com/sourcegraph/scip-typescript/actions/runs/12274624077/workflow

### Test plan
- see if publishing succeeds on main


kritzcreek (approved):
stamp

PR #392, opened 2025-01-07
version 0.3.15

antonsviridov-src:
Tagged release failed, so updating version manually (I deleted the tag and will re-create after this PR)

### Test plan

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sourcegraph/scip-typescript",
-  "version": "v0.3.14",
+  "version": "v0.3.15",
   "description": "SCIP indexer for TypeScript and JavaScript",
   "publisher": "sourcegraph",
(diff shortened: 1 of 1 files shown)


varungandhi-src (approved):
Looks like this didn't publish anything, so this is OK. However, the current CI setup has the unfortunate potential failure mode where you can publish the NPM package successfully, and then fail to publish the Docker image. In that case, deleting the tag + recreating it increases the risk of confusion.

It might be simpler to instead always treat tags as immutable (including for releases which failed to publish), and simply document them as failed in the CHANGELOG.

antonsviridov-src:
> In that case, deleting the tag + recreating it increases the risk of confusion.

Yes, that's the only reason I deleted the tag – because nothing was published.
Agree with treating the tags as immutable.

PR #430, opened 2025-09-26
implemented @deprecated diagnostics reporting

matthewnitschke-wk:
## Problem/Solution

Currently the diagnostics field on the reported index file is always empty, regardless on if the typescript indexer detects any diagnostic issues

This information can be especially helpful analysis of a codebase, where questions like "how many deprecations are being used within this package" are asked. This is something that we rely pretty heavily on in scip-dart use cases (which [fully indexes](https://github.com/Workiva/scip-dart/blob/d1c3a820dfe9a7ef85ce84b52ead40897972d49e/lib/src/metadata.dart#L59) all analyzer results)

This PR updates the scip-typescript indexer to start indexing `@deprecated` jsdoc comments, and assigns the correct `Deprecated` diagnostic tag. Eventually `DiagnosticTag.Unnecessary` should be implemented, but this is less beneficial from my point of view

### Test plan

A new `diagnostics` snapshot test was added. This include updating the snapshot generation to be in line with the results from the official scip cli `snapshot` command: https://github.com/sourcegraph/scip/pull/226


matthewnitschke-wk:
@varungandhi-src this PR should be ready for review. Happy to punt on the idea if you'd prefer the scip-typescript indexer not providing this as well

src/FileIndexer.ts
```diff
@@ -718,6 +719,27 @@ export class FileIndexer {
...
+  ): scip.scip.Diagnostic[] | undefined {
+    const jsDocTags = sym.getJsDocTags()
+
+    const deprecatedTag = jsDocTags.find(tag => tag.name === 'deprecated')
+    if (deprecatedTag) {
+      return [
+        new scip.scip.Diagnostic({
+          severity: scip.scip.Severity.Information,
+          code: 'DEPRECATED',
+          message: deprecatedTag.text?.map(part => part.text).join(''),
```
varungandhi-src, line 751: I don't quite understand when we'll have multiple elements here, and why it's justified to join with `''` instead of say `' '`. Could you explain that? It'd be good to add a code comment for this.

snapshots/output/diagnostics/index.ts
```diff
@@ -0,0 +1,26 @@
...
+//       ^^^ definition diagnostics 0.0.1 `index.ts`/bar().
+//       diagnostic Information:
+//       > This function is deprecated
+
+function main() {
+//       ^^^^ definition diagnostics 0.0.1 `index.ts`/main().
+  new Foo()
+//    ^^^ reference diagnostics 0.0.1 `index.ts`/Foo#
+//    diagnostic Information:
+//    > This class is deprecated
```
varungandhi-src, line 24: It'd be good to add some test cases for multiline messages with `@deprecated` to capture whether they work/don't work, and make sure that the final formatting is correct.

snapshots/output/diagnostics/index.ts
```diff
@@ -0,0 +1,26 @@
+// < definition diagnostics 0.0.1 `index.ts`/
+
+/** @deprecated This class is deprecated */
+class Foo {}
+//    ^^^ definition diagnostics 0.0.1 `index.ts`/Foo#
+//    diagnostic Information:
+//    > This class is deprecated
```
varungandhi-src: I don't have a strong feeling on this (we can change it later), but I'm wondering if it makes sense to emit this at the definition site as well. What does the Dart indexer do? I'm thinking of downstream functionality, and it generally doesn't make sense to flag warnings for definitions that are deprecated, but only at usage sites.

varungandhi-src (changes requested):
Overall I think it's fine to add support for features like this; left some inline comments on specifics.

src/FileIndexer.ts
matthewnitschke-wk: The result of `getJsDocTags()` provides a `JsDocTagInfo` which contains a `text` field on it correlating to a fully tokenized version of the jsdoc

Because the individual parts of this include spaces, we simply `join('')` to generate the full tag text

I've added a comment to explain this better (as well as added a test covering this tokenizing better)

snapshots/output/diagnostics/index.ts
matthewnitschke-wk: Dart omits the diagnostic on the definition side: https://github.com/Workiva/scip-dart/blob/master/snapshots/output/diagnostics/lib/main.dart#L21

I'll update accordingly

snapshots/input/diagnostics/index.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * @deprecated This is a function that has
+ * multiple lines and is also deprecated. Make
+ * sure to reference {@link bar} for some reason
+ */
+function car() {}
+
+function main() {
+  new Foo()
+  bar()
```
varungandhi-src, line 16: Let's add a reference to `car()` to trigger the logic for multiline diagnostics formatting.

varungandhi-src (approved):

PR #471, opened 2026-08-22
Modernize dependencies and tooling

eseliger:
This PR refreshes the repo a bit for 2026:

- update the supported runtime and toolchain to Node.js 22/24 and TypeScript 6 (7 currently does not have the compiler API we rely on, so we aer stuck at 6 for now)
- replace legacy lint/test configuration with ESLint flat config and current tooling
- update runtime dependencies, package-manager fixtures, generated protobuf compatibility, and snapshots
- refresh GitHub Actions, release workflows, documentation, and the pinned Docker image
- remove stale generated build artifacts from fixtures


fkling (approved):

snapshots/output/syntax/src/object-literals-nested.ts
```diff
@@ -30,9 +30,9 @@ export function handleNestedObjectLiterals(): Person {
     address: {
 //  ^^^^^^^ reference syntax 1.0.0 src/`object-literals-nested.ts`/Person#address.
       street: 'Oxford Street',
-//    ^^^^^^ reference syntax 1.0.0 src/`object-literals-nested.ts`/Address#street.
+//    ^^^^^^ definition syntax 1.0.0 src/`object-literals-nested.ts`/street0:
```
christoph-sg: These changes are regressions

snapshots/output/syntax/src/typings.ts
```diff
@@ -3,8 +3,6 @@
 export function process() {
 //              ^^^^^^^ definition syntax 1.0.0 src/`typings.ts`/process().
   return window.process
-//       ^^^^^^ reference typescript 5.6.2 lib/`lib.dom.d.ts`/window.
-//              ^^^^^^^ reference @types/node 20.16.10 `globals.d.ts`/global/process.
-//              ^^^^^^^ reference @types/node 20.16.10 `process.d.ts`/`"process"`/global/process.
+//       ^^^^^^ reference typescript 6.0.3 lib/`lib.dom.d.ts`/window.
```
christoph-sg, line 6: This looks like a regression

snapshots/output/syntax/src/object-literals-arrow-function.ts
```diff
@@ -65,12 +65,12 @@ export function genericArrowOption(): Option<Foobar>[] {
...
 //                                           ^^^^^^ reference syntax 1.0.0 src/`object-literals-arrow-function.ts`/Foobar#
   return [1].map<Option<Foobar>>(n => ({ value: { foobar: n + 1 } }))
-//           ^^^ reference typescript 5.6.2 lib/`lib.es5.d.ts`/Array#map().
+//           ^^^ reference typescript 6.0.3 lib/`lib.es5.d.ts`/Array#map().
 //               ^^^^^^ reference syntax 1.0.0 src/`reusable-types.ts`/Option#
 //                      ^^^^^^ reference syntax 1.0.0 src/`object-literals-arrow-function.ts`/Foobar#
 //                               ^ definition local 22
 //                                       ^^^^^ reference syntax 1.0.0 src/`reusable-types.ts`/Option#value.
-//                                                ^^^^^^ reference syntax 1.0.0 src/`object-literals-arrow-function.ts`/Foobar#foobar.
+//                                                ^^^^^^ definition syntax 1.0.0 src/`object-literals-arrow-function.ts`/foobar0:
```
christoph-sg: A regression

snapshots/output/syntax/src/object-literals-call-signatures.ts
```diff
@@ -65,16 +65,16 @@ export function infersInterface(): void {
     property: [{ value: { property: 42, property2: '42' } }],
 //  ^^^^^^^^ reference syntax 1.0.0 src/`reusable-types.ts`/GenericInterface#property.
 //               ^^^^^ reference syntax 1.0.0 src/`reusable-types.ts`/Option#value.
-//                        ^^^^^^^^ reference syntax 1.0.0 src/`reusable-types.ts`/Numbers#property.
-//                                      ^^^^^^^^^ reference syntax 1.0.0 src/`reusable-types.ts`/Strings#property2.
+//                        ^^^^^^^^ definition syntax 1.0.0 src/`object-literals-call-signatures.ts`/property0:
+//                                      ^^^^^^^^^ definition syntax 1.0.0 src/`object-literals-call-signatures.ts`/property20:
```
christoph-sg: Another regression.

christoph-sg (changes requested):
I spotted a couple regressions in the test output, I'll investigate a bit.

snapshots/output/syntax/src/object-literals-nested.ts
eseliger: I noticed those, I think both versions are correct in some sense, as this is _also_ the definition of a local variable 👀

christoph-sg: You'd definitely want "Find References" on `Adress.street` to find these locations though. It's been a while, but I think TypeScript handles this with a `getContextualType` function on the checker. You didn't change anything about the indexer itself, so it must be a change in behavior on the typescript compiler... I'm looking at it (unless you'd prefer to yourself)

eseliger: be my guest 😀 I won't get to it in the next few days :) Thanks for looking into it!

eseliger: and agreed you'd want this, I think you'd actually want both even (?) so that local go to def to address also works

christoph-sg (approved):
Thank you!

PR #476, opened 2026-08-22
Set SCIP document languages

### LINKED ISSUE #403: Finding No Language, No Kinds in simple node app
[out.txt](https://github.com/user-attachments/files/19567258/out.txt)

Using scip-typescript v0.3.15 on a very simple nodejs project, I took stats off a print I did piping `scip print` into a text file:

   2388           Kind:                   0,
   3401           SyntaxKind:            0,
   14       Language:     "",

It seems not only is the language not being detected, but the Kinds are as well not.

It seems I do get basic info like method names etc though.

Is there some extra I need to add to make this work? I've tried it on several very different repos and doesnt seem to work on any of them.

I attached the print-out created from indexing.

Let me know if I need to add any other details.

eseliger:
Implements support for document language.

Closes https://github.com/sourcegraph/scip-typescript/issues/403.

### DIFF
--- a/src/ProjectIndexer.ts
+++ b/src/ProjectIndexer.ts
@@ -172,7 +172,7 @@ export class ProjectIndexer {
   }
 }
 
-export function languageForFileName(fileName: string): string {
+export function languageForFileName(fileName: string): string | undefined {
   // Document.language uses the exact names from SCIP's Language enum, not the
   // lowercase language identifiers used by editors.
   const extension = path.extname(fileName).toLowerCase()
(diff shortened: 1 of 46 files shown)


src/ProjectIndexer.ts
```diff
@@ -171,6 +172,22 @@ export class ProjectIndexer {
   }
 }
 
+export function languageForFileName(fileName: string): string {
```
christoph-sg: Let's not return empty strings if we can't figure out the language, and return `: string | undefined` instead.

christoph-sg (changes requested):
Small change request. I'll implement it

christoph-sg (approved):

PR #497, opened 2026-09-10
fixes for TS6

christoph-sg:
### Test plan

Updates snapshots

### DIFF
--- a/snapshots/output/syntax/src/object-literals-arrow-function.ts
+++ b/snapshots/output/syntax/src/object-literals-arrow-function.ts
@@ -70,7 +70,7 @@ export function genericArrowOption(): Option<Foobar>[] {
 //                      ^^^^^^ reference syntax 1.0.0 src/`object-literals-arrow-function.ts`/Foobar#
 //                               ^ definition local 22
 //                                       ^^^^^ reference syntax 1.0.0 src/`reusable-types.ts`/Option#value.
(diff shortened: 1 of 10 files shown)


snapshots/output/syntax/src/object-literals-arrow-function.ts
```diff
@@ -70,7 +70,7 @@ export function genericArrowOption(): Option<Foobar>[] {
 //                      ^^^^^^ reference syntax 1.0.0 src/`object-literals-arrow-function.ts`/Foobar#
 //                               ^ definition local 22
 //                                       ^^^^^ reference syntax 1.0.0 src/`reusable-types.ts`/Option#value.
-//                                                ^^^^^^ definition syntax 1.0.0 src/`object-literals-arrow-function.ts`/foobar0:
+//                                                ^^^^^^ reference syntax 1.0.0 src/`object-literals-arrow-function.ts`/Foobar#foobar.
```
eseliger, line 73: this is definitely good because it restores the original behavior, but I wonder if this shouldn't emit both a definition and this reference? 🤔

eseliger (approved):
Awesome thanks!

snapshots/output/syntax/src/object-literals-arrow-function.ts
christoph-sg: As TypeScript's type system is structural, the whole definition/reference distinction is best effort... I think in this case the reference is justified, as the explicit type annotation of `FooBar`, makes it clear that's the `foobar` field the user is talking about here.

